PR #1, opened 2021-08-19
Proposal for probeye's first version

aklawonn:
As the title suggests, this is a proposal for the initial version of probeye. The pull request should give all contributors the chance to look on the code, and comment on what they would like to change before the merge into the main-branch.


probeye/forward_model.py
```diff
@@ -0,0 +1,251 @@
...
+        self.output_sensors = output_sensors
+
+    def __call__(self, exp_inp, prms):
+        """
+        Evaluates the model response for each output sensor and returns the
+        response dictionary.
+
+        Parameters
+        ----------
+        exp_inp : dict
```
joergfunger, line 98: As mentioned above, I would call that data that then includes deterministic as well as measured quantities.

probeye/forward_model.py
```diff
@@ -0,0 +1,251 @@
...
+            numeric values as values.
+
+        Returns
+        -------
+        response_dict : dict
+            Contains the model response (value) for each output sensor,
+            referenced by the output sensor's name (key).
+        """
+        inp = {**exp_inp, **prms}  # adds the two dictionaries
+        response_dict = {output_sensor.name: self.response(inp, output_sensor)
```
joergfunger, line 112: I would require the user to implement this function (instead of having a response function).

probeye/forward_model.py
```diff
@@ -0,0 +1,251 @@
...
+            prms_right[prm_name] = prm_value + h
+            right = self(exp_inp, prms_right)
+            # evaluate the symmetric difference scheme
+            jac[prm_name] = {}
+            for sensor_name in left.keys():
+                jac[prm_name][sensor_name] =\
+                    (right[sensor_name] - left[sensor_name]) / (2 * h)
+        return jac
+
+    def error_function(self, exp_inp, prms, output_sensor, ye):
```
joergfunger, line 174: this could be  done by the metric_function at the sensor.

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+
+        # this dictionary is intended for storing the measured data from
+        # experiments (see self.add_experiment); this dict is managed
+        # internally and should not be edited directly
+        self._experiments = {}
+
+        # here, the available prior classes are stored; additional ones can be
+        # added via self.add_prior_class; see also priors.py; this dictionary is
+        # managed internally and should not be edited directly
+        self._prior_classes = {'normal': PriorNormal,
```
joergfunger, line 63: What is the advantage of using a string vs directly specififying a base class that can then be directly added by the user (without changing the library code)?

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+        self._noise_models = {}
+
+    @property
+    def n_prms(self):
+        """Provides n_prms attribute."""
+        return self._parameters.n_prms
+
+    @property
+    def n_calibration_prms(self):
+        """Provides n_calibration_prms attribute."""
```
joergfunger, line 93: What's the difference, is the first one the deterministic + latent, whereas the second is only the latent ones?

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+                      info="No explanation provided", tex=None):
+        """
+        Adds a parameter ('const' or 'calibration') to the inference problem.
+
+        Parameters
+        ----------
+        prm_name : string
+            The name of the parameter which should be added to the problem.
+        prm_type : string
+            Either 'model' (for a model parameter), 'prior' (for a prior
```
joergfunger, line 203: this is not fully clear to me. Why shouldn't I define a hyperparameter of the prior not via this definition? I would have thought to first define the hyper prior, and then use that variable as a definition for the model parameter prior. So from my point of view, variables are either model, prior or noise (X0, X1, X2) or constant.

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+            parameter) or 'noise' (for a noise parameter). Note that 'prior'
+            parameter do not have to be added by hand. They are added
+            internally when 'calibration'-parameters are added.
+        const : float or None, optional
+            If the added parameter is a 'const'-parameter, the corresponding
+            value has to be specified by this argument.
+        prior : tuple of two elements or None, optional
+            If the added parameter is a 'calibration'-parameter, this argument
```
joergfunger, line 211: calibration parameter = latent variable?

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+        if const is None:  # in this case we are adding a 'calibration'-param.
+            # first, define the index of this parameter in the numeric vector
+            # theta, which is given to self.loglike and self.logprior
+            prm_index = self._parameters.n_calibration_prms
+            # the prm_value is reserved for 'const'-parameter; hence, it is set
+            # to None in this case, where we are adding a 'calibration'-param.
+            prm_value = None
+            # the remaining code in this if-branch defines the prior that is
+            # associated with this 'calibration'-parameter
+            prior_type = prior[0]  # e.g. 'normal', 'lognormal', etc.
```
joergfunger, line 263: As mentioned above, I feel like a class would be better here. Then everyone can directly define the parameters of the prior (so Gamma might have alpha, and normal mu and sigma).

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+            del self._parameters[prm_name].prior
+            del self._parameters[prm_name]
+            # correct the indices of the remaining 'calibration'-parameters
+            idx = 0
+            for name, parameter in self._parameters.items():
+                if parameter.index is not None:
+                    parameter.index = idx
+                    idx += 1
+
+    def change_parameter_role(self, prm_name, const=None, prior=None,
```
joergfunger, line 344: For me, the need for this method is rather unclear. I would directly define that in the definition, and potentially having to define the outer variables first. (So first define the hyperparameters prior, and then define the model parameters prior with parameters that are already latent variables).

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+                    f"The noise model parameter '{prm_name}' has not been " +
+                    f"defined yet.\nYou have to add all noise model " +
+                    f"parameters to the problem before adding the noise " +
+                    f"model.\nYou can use the 'add_parameter' method for " +
+                    f"this purpose."
+                )
+
+        # add the given noise model to the internal noise model dictionary under
+        # the given name of the output sensor(s)
+        self._noise_models[output_sensor_name] = noise_model
```
joergfunger, line 847: The noise model does not necessarily relate to the sensor (e.g. if you have a correlated noise model with stereo camera data and laser measuring the same thing, thus are correlated),you would have to define a "joint" noise model.

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+
+    def logprior(self, theta):
```
joergfunger, line 849: those functions should be moved to the inference, since the computation (e.g.the evaluation of the likehood or prior) is done there.

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+        lp : float
+            The evaluated log-prior function for the given theta-vector.
+        """
+        lp = 0.0
+        for prior in self._priors.values():
+            prms = self.get_parameters(theta, prior.prms_def)
+            lp += prior(prms, 'logpdf')
+        return lp
+
+    def sample_from_prior(self, prm_name, size):
```
joergfunger, line 871: see above

probeye/noise.py
```diff
@@ -0,0 +1,79 @@
...
+            A list of parameter names (strings) defining how a noise parameter
+            vector given to the loglike_contribution method is interpreted. E.g.
+            prms_def = ['mu', 'sigma'] means that the noise parameter vector
+            has two elements, the first of which gives the value of 'mu' and the
+            second gives the value of 'sigma'.
+        """
+        self.prms_def = list2dict(prms_def)
+        self.prms_dim = len_or_one(prms_def)
+
+    def loglike_contribution(self, model_error_vector, prms):
```
joergfunger, line 27: it is not necessarily the model error, but the output of the sensor_metric function.

probeye/parameter.py
```diff
@@ -0,0 +1,61 @@
+class Parameters(dict):
+    """
+    The main parameter 'library'. In this dictionary, all of the problem's
```
joergfunger, line 3: Should we consider extending this to vector parameters?

probeye/parameter.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        Parameters
+        ----------
+        prm_dict : dict
+            The keys are 'index', 'type', 'role', 'prior', 'value', 'info' and
+            'tex', while the values are the corresponding values of these
+            properties. See also the explanations in InferenceProblem.__init__()
+            for more detailed information.
+        """
+        self.index = prm_dict['index']
```
joergfunger, line 55: for vector parameters, we would add the length

probeye/priors.py
```diff
@@ -0,0 +1,272 @@
...
+from scipy import stats
+
+# local imports
+from probeye.subroutines import list2dict
+
+
+class PriorTemplate:
+    """
+    Template class for prior definitions. Note that the main motivation of how
+    this class is defined is to avoid storing any numeric values for any of the
```
joergfunger, line 11: delete is defined

probeye/priors.py
```diff
@@ -0,0 +1,272 @@
...
+        -------
+        float
+            The result of stats.norm.<method>(x, loc, scale).
+        """
+        x = prms[self.ref_prm]
+        loc = prms[f"loc_{self.ref_prm}"]
+        scale = prms[f"scale_{self.ref_prm}"]
+        return getattr(stats.norm, method)(x, loc, scale)
+
+    def generate_samples(self, prms, size, seed=None):
```
joergfunger, line 83: is that something that is a description of the inference problem or that is a helper function in the inerence? From my point of view, this is the latter and then might be moved there.

probeye/solver.py
```diff
@@ -0,0 +1,72 @@
...
+import matplotlib.pyplot as plt
+
+# local imports
+from taralli.parameter_estimation.base import EmceeParameterEstimator
+
+# ============================================================================ #
+#                           Taralli sampling solver                            #
+# ============================================================================ #
+
+def taralli_solver(problem, n_walkers=20, n_steps=1000, plot=True,
```
joergfunger, line 16: move this solver to a subdirectory solver/taralli

joergfunger (changes requested):
In the first review, I did not have time to look into the test cases. But I guess with the changes tey might change as well. In general, this is a very promising first version, in particular the documentation is very nice.

examples/example_linear_regression.py
```diff
@@ -0,0 +1,121 @@
...
+                      prior=('uniform', {'low': low_sigma, 'high': high_sigma}))
+
+# add the forward model to the problem
+inp_1 = InputSensor("x")
+out_1 = OutputSensor("y")
+problem.add_forward_model("LinearModel",
+                          LinearModel([{'a': 'm'}, 'b'], [inp_1], [out_1]))
+
+# add the noise model to the problem
+problem.add_noise_model(out_1.name, NormalNoiseZeroMean(['sigma']))
```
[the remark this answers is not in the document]
aklawonn: 'sigma' is supposed to be identified here. This can be checked by looking at the definition of 'sigma' when it was added to the problem. Since it is not a constant (it was assigned a prior) it will be inferred.

probeye/parameter.py
aklawonn: We can do that, but I guess it would require changes at several locations in the code. I would suggest to add this change after we have published the initial version, since it is IMO more of a special case when you have vector-type parameters.

probeye/parameter.py
aklawonn: Yes, for that case, we would need to add a length-attribute. As suggested before, I would implement that change, after publishing the initial version, if you don't mind.

examples/example_linear_regression.py
```diff
@@ -0,0 +1,121 @@
...
+
+# ============================================================================ #
+#                         Define the Inference Problem                         #
+# ============================================================================ #
+
+# initialize the inference problem with a useful name
+problem = InferenceProblem("Linear regression with normal noise")
+
+# add all parameters to the problem
+problem.add_parameter('a', 'model',
```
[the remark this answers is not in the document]
aklawonn: I added more explanations to the code in the linear regression example. I will add it with the next push to this branch.

examples/example_linear_regression.py
```diff
@@ -0,0 +1,121 @@
...
+                      prior=('normal', {'loc': loc_b, 'scale': scale_b}))
+problem.add_parameter('sigma', 'noise',
+                      info="Std. dev, of 0-mean noise model", tex=r"$\sigma$",
+                      prior=('uniform', {'low': low_sigma, 'high': high_sigma}))
+
+# add the forward model to the problem
+inp_1 = InputSensor("x")
+out_1 = OutputSensor("y")
+problem.add_forward_model("LinearModel",
+                          LinearModel([{'a': 'm'}, 'b'], [inp_1], [out_1]))
```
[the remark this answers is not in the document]
aklawonn: Yes indeed. I added some comments on that. Will be added with the next push.

examples/example_linear_regression.py
```diff
@@ -0,0 +1,121 @@
...
+# data-generation process; normal noise with constant variance around each point
+np.random.seed(seed)
+x_test = np.linspace(0.0, 1.0, n_tests)
+y_true = a_true * x_test + b_true
+y_test = np.zeros(n_tests)
+for i in range(n_tests):
+    y_test[i] = np.random.normal(loc=y_true[i], scale=sigma_noise)
+
+# add the experimental data
+problem.add_experiment(f'TestSeries_1', sensors={'x': x_test, 'y': y_test},
```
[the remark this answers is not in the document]
aklawonn: I like the term sensor in this context, because it mirrors the setup of the forward model. If you have a sensor with name 'x' in the forward model, there has to be a sensor called 'x' in the experiment (and the noise model will take these two sensors, and compare them). I think this concept is the easiest to grasp, when both data containers are called 'sensor'. I can see, that the term data would be appealing in that sense, that it is more neutral, but I would think the mirror-concept is more valuable to have.

examples/example_linear_regression.py
```diff
@@ -0,0 +1,121 @@
...
+plt.legend()
+plt.tight_layout()
+plt.draw()  # does not stop execution
+
+# ============================================================================ #
+#                          Solve problem with Taralli                          #
+# ============================================================================ #
+
+# code is bundled in a specific solver routine
+taralli_solver(problem, n_walkers=n_walkers, n_steps=n_steps)
```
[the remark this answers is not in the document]
aklawonn: Maybe you just looked on the code in your browser, because if you run this example, some post-processing is happening, and two plots are created.

examples/example_multiple_sensors.py
```diff
@@ -0,0 +1,153 @@
...
+
+# taralli settings
+n_walkers = 20
+n_steps = 2000
+
+# ============================================================================ #
+#                           Define the Forward Model                           #
+# ============================================================================ #
+
+class PositionSensor(OutputSensor):
```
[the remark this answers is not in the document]
aklawonn: I would argue, that any numeric datum in the experiments is provided by some sensor, because it was measured in one way or the other. so there is only sensors. but in the forward model, which mirrors some sensors of the experiments, some of those sensors are interpreted as input and some as output. simply because the forward model is a function with a domain (input) and an image (output). I would not say, that every sensor is an output sensor.

examples/example_multiple_sensors.py
```diff
@@ -0,0 +1,153 @@
...
+                      prior=('uniform', {'low': low_S2, 'high': high_S2}),
+                      info="Std. dev. of zero-mean noise model for S1",
+                      tex=r"$\sigma_2$")
+problem.add_parameter('sigma_3', 'noise',
+                      prior=('uniform', {'low': low_S3, 'high': high_S3}),
+                      info="Std. dev. of zero-mean noise model for S1",
+                      tex=r"$\sigma_3$")
+
+# add the forward model to the problem
+inp_1 = InputSensor("time")
```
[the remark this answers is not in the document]
aklawonn: The latter (i.e. specify the time steps as a bunch of parameters, or at least prepare the possibility to do so) is IMO too confusing for a simple example. Regarding the sensors, maybe we should have a discussion on the general approach here, because there are a lot of comments on them. For the reasons mentioned in previous responses, I would like to keep the name InputSensor.

examples/example_multiple_sensors.py
```diff
@@ -0,0 +1,153 @@
...
+
+# give problem overview
+problem.info()
+
+# ============================================================================ #
+#                          Solve problem with Taralli                          #
+# ============================================================================ #
+
+# code is bundled in a specific solver routine
+taralli_solver(problem, n_walkers=n_walkers, n_steps=n_steps)
```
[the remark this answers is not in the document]
aklawonn: Similar to a comment above: post-processing is happening here within the taralli solver. A result plot is generated when running the script.

examples/example_linear_regression.py
```diff
@@ -0,0 +1,121 @@
...
+# ============================================================================ #
+#                           Define the Forward Model                           #
+# ============================================================================ #
+
+class LinearModel(ModelTemplate):
+    def response(self, inp, sensor):
+        x = inp['x']
+        a = inp['m']
+        b = inp['b']
+        return a * x + b
```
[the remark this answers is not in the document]
aklawonn: Valid point. I opened an issue on that matter.

probeye/forward_model.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    Base class for an input sensor of the forward model. Essentially just a
+    class with a 'name' attribute.
+    """
+    def __init__(self, name):
+        """
+        Parameters
+        ----------
+        name : string
+            The name of the sensor, e.g. 'Deflection-Sensor bottom-left'.
```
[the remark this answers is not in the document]
aklawonn: I think you can see the deflection as an input, it the experiment is deflection-controlled, but I guess it's more clear for another example. I changed it to 'Time'.

probeye/forward_model.py
```diff
@@ -0,0 +1,251 @@
...
+        error_metric : string, optional
+            Either 'abs' (absolute) or 'rel' (relative). Defines if the model
+            error with respect to this sensor should be measured in absolute or
+            relative terms. In the former case ('abs'), the error is defined as
+            (model prediction - measured value). In the latter case ('rel') the
+            error is defined as (1 - (model prediction / measured value)).
+        """
+        self.name = name
+        self.error_metric = error_metric
```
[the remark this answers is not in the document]
aklawonn: I think if a relative error makes sense, also the experimental datum should not become zero. I like the definitions "model - measured" and "1 - model/measured" because positive/negative numbers mean, that your model error is positive/negative. In the other way of defining it the meaning flips, which is a bit confusing to me.

probeye/forward_model.py
```diff
@@ -0,0 +1,251 @@
...
+            contain strings or one-element dictionaries. It could look, for
+            example, like [{'a': 'm'}, 'b']. The one-element dictionaries
+            account for the possibility to define a local name for a calibration
+            parameter that is different from the global name. In the example
+            above, the calibration parameter with the global name 'a' will be
+            referred to as 'm' within the model. So, the one-element dicts have
+            the meaning {<global name>: <local name>}. String-elements are
+            interpreted as having similar local and global names. Note that the
+            local-name option will not be required most of the times.
+        input_sensors : list[InputSensor]
```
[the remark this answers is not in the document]
aklawonn: I think this distinction between input and parameters is quite common (see for example https://en.wikipedia.org/wiki/Uncertainty_quantification where the input is called 'x' and the parameters 'theta'). This is distinction is necessary, for example when it comes to computing the jacobian.

probeye/forward_model.py
aklawonn: See responses above. I would argue there are no deterministic values in real-world data, since all values come from measurements.

probeye/forward_model.py
aklawonn: This is probably solved when issue 3 is solved.

probeye/forward_model.py
aklawonn: I thought about it, but I forgot why I didn't do it. Certainly worth to look into. I opened an issue on that front.

probeye/inference_problem.py
aklawonn: You can add your own prior-class via InferenceProblem.add_prior_class.

probeye/inference_problem.py
aklawonn: 'calibration'-parameters are the parameters that are supposed to be inferred. 'const'-parameters have a fixed value, and are not inferred.

probeye/inference_problem.py
aklawonn: 'model', 'noise' and 'prior' are the parameter types. 'calibration' and 'const' are the parameter roles. each parameter of any type can have the role 'calibration' (when it should be inferred) or 'const' (when it should not be inferred). Prior parameters are always added automatically with the role 'const', when adding a 'calibration'-parameter to the problem. if a prior-parameter should not be 'const', you can change its role to 'calibration' with the method InferenceProblem.change_parameter_role.

probeye/inference_problem.py
aklawonn: Yes, I chose that name, because the term 'latent variable' is often used when comparing it to 'observable variables'. Here, however, the other group is not 'observable variables' but simply 'constants'. Also, I thought the term 'calibration'-parameter is easier to understand for someone who wants to use probeye, but isn't too familiar with calibration problems.

probeye/inference_problem.py
aklawonn: Maybe there's a better way to do it as it is now. But you can add your own prior class with InferenceProblem.add_prior_class.

probeye/inference_problem.py
aklawonn: Good point, correlations cannot be defined yet. I guess this is something for a future version though.

probeye/priors.py
aklawonn: Well I can't just delete 'is defined' here, because then the sentence doesn't make any sense anymore. Instead, I changed 'defined is' to 'implemented was'.

probeye/noise.py
aklawonn: Aren't we calling the output of the sensor_metric function 'model error' so far?

probeye/subroutines.py
```diff
@@ -0,0 +1,174 @@
...
+    if delta is not None:
+        return delta
+    eps = 1e-7  # approx sqrt(machine precision)
+    dx = x0 * eps + eps  # TODO: clarify if the '+ eps' part is correct
+    if dx == 0:
+        dx = eps
+    return dx
+
+def underlined_string(string, symbol="=", n_empty_start=1, n_empty_end=1):
+    """
```
TTitscher, line 53: Suggestion: See if
~~~py
def underlined_string(string)
    return tabulate(None, headers=[string], tablefmt="grid")
~~~
looks good enough for you.

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
+# standard library imports
+from copy import deepcopy as dc
```
TTitscher, line 2: I was a bit confused when i saw `dc` in the code and would find `copy.deepcopy` more explicit.

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+    def n_calibration_prms(self):
+        """Provides n_calibration_prms attribute."""
+        return self._parameters.n_calibration_prms
+
+    @property
+    def n_constant_prms(self):
+        """Provides n_constant_prms attribute."""
+        return self._parameters.n_constant_prms
+
+    def info(self, print_it=True, include_experiments=False):
```
TTitscher, line 101: I recommend the `tabulate` package to print this kind of info. With the option `tablefmt="latex"`, you could directly put that into a paper or presentation. Also, it would get rid of the string formatting functions in `subroutines`.

probeye/parameter.py
```diff
@@ -0,0 +1,61 @@
...
+        self.type = prm_dict['type']
+        self.role = prm_dict['role']
+        self.prior = prm_dict['prior']
+        self.value = prm_dict['value']
+        self.info = prm_dict['info']
+        self.tex = prm_dict['tex']
```
TTitscher, line 61:
```suggestion
        self.__dict__ = prm_dict
```
would do the same.

probeye/inference_problem.py
```diff
@@ -0,0 +1,934 @@
...
+        w = len(max(prms_roles_types['const'], key=len)) + 2
+        for prm_name in prms_roles_types['const']:
+            prm_value = self._parameters[prm_name].value
+            const_prms_str += tcs(prm_name, f"{prm_value:.2f}", col_width=w)
+
+        # additional information on the problem's parameters
+        prms_info_str = underlined_string("Parameter explanations", symbol="-")
+        w = len(max(self._parameters.keys(), key=len)) + 2
+        for prm_name, parameter in self._parameters.items():
+            prms_info_str += tcs(prm_name, f"{parameter.info}", col_width=w)
```
TTitscher, line 155: I would think about if anything related to the parameters should go into a `Parameters.info(...)` to be called separately. (Also having a LaTeX parameter-table for a model overview in mind...)

probeye/solver.py
aklawonn: I moved it.

probeye/priors.py
aklawonn: Do you mean the generate_samples method? I think this is a general method which just belongs to the prior-class. I mean, strictly speaking, also the __call__ method (which evaluates the logpdf) is not describing the inference problem, but just a helper function in the inference. Maybe we can have a chat on that matter.

TTitscher (changes requested):
Thanks for putting that together and also adding tests, nice!

As commented in the `ModelTemplate` code, I just realized that the `__call__` method is called once **per output sensor** instead of really just once. That needs to be changed and I'll approve the PR as soon as this is fixed.

Also the recommendation to overwrite the `reponse` method may be troublesome as this makes separating an expensive calculation from a cheap postprocessing (to just extract the response for a given sensor) not trivial.

Apart from that I added some suggestions here and there for you to consider.

probeye/inference_problem.py
aklawonn: This is probably a good idea, however, right now, I don't really see a place where else to put them. I will have to think about this a bit. I would suggest, for the first version, we leave it as is. Otherwise, the structure might change considerably to what was presented in the meeting where we first presented probeye.

probeye/inference_problem.py
aklawonn: See response above.

probeye/parameter.py
aklawonn: You're right, that's much shorter. But then you could essentially add any dictionary to ParameterProperties without any checks if all required keys are there. Of course, you could add some checks then, but I thought, the way it is implemented right now, would just throw an error if a required key does not appear. Also, I thought it is easier to see at one glance which attributes this class has in the way it is implemented now compared to the short version.

probeye/inference_problem.py
aklawonn: No problem, I changed it.

examples/example_two_models.py
```diff
@@ -0,0 +1,162 @@
...
+
+# ============================================================================ #
+#                    Add test data to the Inference Problem                    #
+# ============================================================================ #
+
+# data-generation process; normal noise with constant variance around each point
+np.random.seed(seed)
+x_test = np.linspace(0.0, 1.0, n_tests)
+y_linear_true = a_true * x_test + b_true
+y_quadratic_true = alpha_true * x_test**2 + b_true
```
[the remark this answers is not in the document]
aklawonn: Good point, done.

probeye/priors.py
```diff
@@ -0,0 +1,272 @@
...
+
+        Returns
+        -------
+        float
+            The result of stats.norm.<method>(x, loc, scale).
+        """
+        x = prms[self.ref_prm]
+        loc = prms[f"loc_{self.ref_prm}"]
+        scale = prms[f"scale_{self.ref_prm}"]
+        return getattr(stats.norm, method)(x, loc, scale)
```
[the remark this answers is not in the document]
aklawonn: Your first point is a good idea, I changed that. Regarding the meta-programming, I wanted to avoid it since it is rather slow.

examples/example_linear_regression.py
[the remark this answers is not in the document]
joergfunger: Yes, I did. But I guess it would be best to move this a separate postprocessing call instead of hiding that in the wrapper class for Taralli. Maybe something to be done in a subsequent step and not in this first version, in that case could you add an issue?

probeye/priors.py
joergfunger: yes, and that is why I think it is not part of the description of the inference problem, but already part of the solution, which should be moved to the inference engine

examples/example_multiple_sensors.py
```diff
@@ -0,0 +1,153 @@
...
+    time_steps = np.linspace(0, 1, n_time_steps)
+    prms = {'A': A_true, 'B': B_true}
+    sensors = linear_model({'time': time_steps}, prms)
+    for key, val in sensors.items():
+        sensors[key] = val +\
+                       np.random.normal(0.0, sd_dict[key], size=n_time_steps)
+    sensors['time'] = time_steps
+    problem.add_experiment(f'TestSeries_{n}',
+                           sensors=sensors,
+                           fwd_model_name='LinearModel')
```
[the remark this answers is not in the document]
joergfunger: Getting rid of the sensor object makes it difficult to store information e.g. at what position the sensor is located. For me, the sensor is just a construct to define what is computed in a postprocessing step of the forward model (so that you can essentially decide what data to be used without changing the forward model). An input sensor is for me less intuitive.

probeye/priors.py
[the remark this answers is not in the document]
joergfunger: IMO, this evaluation part should be moved into the inference engine and then everyone can decide how to implement the evaluation of the pdfs (using scipy, pyro, ..)

examples/example_multiple_sensors.py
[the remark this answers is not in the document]
joergfunger: For me, the (output)-sensor in the forward model defines what is to be returned. The input is not necessarily something that is related to an experiment, but could be e.g. the mesh_size used. For me, there is no clear distinction between what is an input and what is a parameter, maybe apart from the fact that the first ones are static (so you could create a response surface with only the parameters being variable), or you could e.g. create the FEM mesh with these parameters (e.g. the mesh) and then only solve for this constant mesh. As such, I feel the forward modul should get variable/latent and static parameters (the latter could even be provided in the constructor or some init function before the actual computation is performed e.g. to create the mesh, factorize the stiffness matrix, compile the fenics form, ..) and then only have inputs (latent) and sensors (outputs).

probeye/forward_model.py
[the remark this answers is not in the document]
joergfunger: valid point, in a displacement-controled setting the displacement is an input. Even though I think that should not be called sensor, but i suggest that we have (as you already discussed in the previous responses) an online meeting on that on what we think is a sensor/input/output/data,..

probeye/forward_model.py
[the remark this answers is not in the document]
joergfunger: as for the relative definition of the error there is certainly not a right or wrong. Related to the error_metric, we should probably discuss where this should be implemented (e.g. at the noise model, the sensor) and how (as a function, as a string).

probeye/forward_model.py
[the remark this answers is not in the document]
joergfunger: see my comments above, thus the x parameters can be provided a priori and do not have to be part of the call function, because they are always constant (and allow you to appropriately set up the model).

probeye/forward_model.py
joergfunger: From the modeler perspective, it does not make a difference whether the input is from the experiment, or if it is prescribed by the one doing the inference. What matters IMO is whether input data is constant/deterministic or variable/latent. I realize that exp. input might be determinstic, but still can't be stored in an init function if the same forward model is used for multiple datasets (with potentially different deterministic values). But maybe it makes sense to exclude that option so that people would have to create multiple forward models with different deterministic inputs.

probeye/inference_problem.py
joergfunger: so why do we need to store these prior_classes in the inference problem, I would argue we just need the priors and then instantiate an object of each prior type (e.g. PriorNormal?)

probeye/inference_problem.py
joergfunger: maybe we should develop a clear wording, because we sometimes talk about latent, variable or calibration parameters often refering to the same concept (or sometimes maybe not:)

probeye/inference_problem.py
joergfunger: "In statistics, latent variables (from Latin: present participle of lateo (“lie hidden”), as opposed to observable variables) are variables that are not directly observed but are rather inferred" from [wikipedia](https://en.wikipedia.org/wiki/Latent_variable). I would interpret that as the calibration parameters (because we can't observe them). But maybe that is a point for the next meeting as well.

probeye/inference_problem.py
joergfunger: Why wouldn't you just change the defition of the parameters, instead of first defining them as variable and then afterwards redefining them as const? As for the prior-prior option, you can define the hyperparameter with its prior first (then it is automatically defined as being latent, and only afterwards define the prior of a model parameter with this hyperparameter.

probeye/noise.py
joergfunger: Here it also seems to be necessary to define the wording. For me, the model error is the difference between exp data and forward model (without applying some metric to make it relative/abs or any other transformation). But we just have to agree on the wording.

examples/example_multiple_sensors.py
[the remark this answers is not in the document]
TTitscher: As I said, I understand the concept of a sensor as a nice way to connect the model and the data. But I see little/no benefit from using/subclassing these `Input/OutputSensor` classes, when I could also use a string, a number, a custom `Sensor` class, a dictionary with inputs, ... 

But maybe having these classes makes it easier for users to follow the sensor *concept*.

probeye/subroutines.py
aklawonn: I looked at what you started with tabulate before, and revised the InferenceProblem's info method accordingly.

probeye/inference_problem.py
aklawonn: Done, now the info-printout (more specifically, the tables appearing in it) is constructed with tabulate.

probeye/inference_problem.py
aklawonn: I implemented that. Now, the Parameter's class has some info-methods using tabulate.

examples/example_linear_regression.py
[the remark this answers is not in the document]
aklawonn: All right, I opened an issue on that matter.

probeye/forward_model.py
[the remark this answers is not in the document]
aklawonn: The x-values cannot be provided a priori, because they differ from experiment to experiment. They are the 'observable variables' while theta contains the 'latent variables'.

probeye/forward_model.py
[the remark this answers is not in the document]
aklawonn: Right, I think this comment is covered by the issue on the error-metric.

probeye/inference_problem.py
aklawonn: Solid point, I opened an issue on that matter.

probeye/forward_model.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        inp = {**exp_inp, **prms}  # adds the two dictionaries
+        response_dict = {output_sensor.name: self.response(inp, output_sensor)
+                         for output_sensor in self.output_sensors}
+        return response_dict
+
+    def response(self, inp, sensor):
+        """
+        Evaluates the model for the given sensor. This method has to be
+        overwritten by the user's response method.
```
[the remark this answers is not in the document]
aklawonn: You're right, I looked into it, and revised the __call__ method of ModelTemplate accordingly.

probeye/forward_model.py
```diff
@@ -0,0 +1,251 @@
...
+            The experimental output value for the given output_sensor.
+
+        Returns
+        -------
+        error : float
+            The computed model error for the given output sensor.
+        """
+
+        # compute the model prediction for the given sensor
+        ym = self(exp_inp, prms)[output_sensor.name]
```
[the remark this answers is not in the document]
aklawonn: Yep, that implementation wasn't a bright star in the efficiency-sky. I worked that over.

TTitscher (approved):

examples/example_multiple_sensors.py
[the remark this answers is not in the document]
aklawonn: I see what you understand under a sensor in this context. However, I think that this understanding is not very intuitive. For you, a sensor is essentially an outward directed data-channel from some computed result by a forward model. But when I think of a sensor, I think of a physical device that measures something. That's why I find it very intuitive to associate a sensor class with the data from an experiment - because all this data was measured by physical sensors. The forward model now takes some of these sensor-measurements as input. Hence, for the forward model sees these sensors are input sensors. And the values of the input sensors are not static, since they vary from experiment to experiment. The inputs you are mentioning (e.g. the mesh-size) are just constants, i.e. constant parameters of the forward models. These can either be defined within the forward model without introducing them in the inference problem, or they can just be defined as constants in the inference problem. The forward model in the current implementation has also output sensors how you understand them. But in my understanding they are just abstract concepts that mirror the actual sensors from the experimental setup.

probeye/forward_model.py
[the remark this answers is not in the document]
joergfunger: I would say the observable variables are usually the outputs, but I realize that there are deterministic parameters that are constant for each model (for all experiments) and others that are constant but depend on the experiment. So we do not distinguish between those two, so inputs are the deterministic parameters, whereas prms are the ones we are calibrating.

examples/example_linear_regression.py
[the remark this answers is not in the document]
aklawonn: I separated the taralli solver routine from the postprocessing. Also, there is no taralli.py or taralli subpackage anymore, so there shouldn't be any import-clashes anymore.

examples/example_linear_regression.py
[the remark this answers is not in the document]
aklawonn: By now, this problem, which was also mentioned by @TTitscher has been resolved. The corresponding issue #3 was closed.

probeye/forward_model.py
aklawonn: Issue #3 was closed, so this point here should be covered by now.

probeye/inference_problem.py
aklawonn: Issue #6, which was dealing with this point was closed. The prior classes have been decoupled from the InferenceProblem.

probeye/inference_problem.py
aklawonn: I restructured the package into a definition and an inference subpackage. In the definition package, there are no computing routines anymore. All computing routines are moved to the inference subpackage.

probeye/priors.py
aklawonn: The separation into a definition and an inference subpackage, where all computing routines have been moved to the inference subpackage resolved this point.

probeye/priors.py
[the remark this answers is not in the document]
aklawonn: The prior's computing routines have been moved to the inference subpackage and are not part of the problem definition anymore.

probeye/inference_problem.py
aklawonn: One can now directly define prior-parameters as being calibration parameters without utilizing change_parameter_role (see test_prior_calibration.py for an example). However, the method change_parameter_role is still usefull to switch parameters from const to calibration (latent) or vice versa.

probeye/definition/forward_model.py
```diff
@@ -0,0 +1,193 @@
...
+            # evaluate the model at prms_i + h
+            prms_right = cp.copy(prms)
+            prms_right[prm_name] = prm_value + h
+            right = self({**exp_inp, **prms_right})
+            # evaluate the symmetric difference scheme
+            jac[prm_name] = {}
+            for sensor_name in left.keys():
+                jac[prm_name][sensor_name] =\
+                    (right[sensor_name] - left[sensor_name]) / (2 * h)
+        return jac
```
atulag0711, line 127: Hi, Alex,
Thanks for the well structured and commented code. Its super sleek and I think very comprehensive. 
About the Jacobian above, wouldn't that be beneficial to have an option to check if the forward solver passes adjoint/adoint operator. If so then that should be returned by the Jacobian operator. If the solver is not differentiable then only CD based Jacobian needs to be passed. Just an idea.

atulag0711: Apart from this I think our video call clarified a lot of stuff. waiting for 1.0 then Ill build a pyTorch wrapper around this.

joergfunger: We have to distinguish between a jacobian (the derivative of all model outputs with respect to all parameters = matrix) and a jacobian of the loglikelihood with respect to the parameters (just a vector).

probeye/inference/taralli_/postprocessing.py
```diff
@@ -0,0 +1,39 @@
...
+
+    # plot the results
+    if plot:
+        tex_labels = problem.get_theta_names(tex=True)
+        emcee_model.plot_posterior(dim_labels=tex_labels)
+        plt.show(block=True)
+
+    # prints a summary to the console
+    if summary:
+        emcee_model.summary()
```
atulag0711, line 39: IMHO, there can be solver specifi viz but there should be a common viz for all taralli/Torch/tensorflow ..right. Ultimately whatever we use we have posterior samples in array or form of  approximate distribution. (VI)

probeye/inference/taralli_/priors.py
```diff
@@ -0,0 +1,313 @@
...
+    # prepare the corresponding prior object; the following translation is
+    # necessary, because prms_def must be given in form of a list, but was
+    # already translated to a dictionary when instantiating the PriorTemplate
+    # objects; hence prior_template.prms_def is a dictionary
+    prms_def = [{key: value} for key, value in prior_template.prms_def.items()
+                if key != prior_template.ref_prm]
+    prior_object = prior_classes[prior_template.prior_type](
+        prior_template.ref_prm, prms_def, prior_template.name)
+
+    return prior_object
```
atulag0711, line 313: Like priors, noise model will also be solver specific. For eg torch will need torch distributions

probeye/definition/forward_model.py
atulag0711: @joergfunger Hi Jorg. Yes valid point. I think if dL/d\theta  = dL/dy * dy/d\theta.  dy/d\theta is a matrix and it comes from the forward solve which needs to be stored somewhere (also updated after every solve). dL/dy is computed by autodiff (e.g torch).  These two gives us dL/d\theta which is needed to move to next step in the optimizer. Please correct me if I am wrong/unclear.

Implementation wise, the Jacobian of the solver (dy/d\theta) can be an operator (FEnics) and if supplied by the likelihood function should be able to give adjoints.

joergfunger: The difference is that we can always compute dL/dtheta if we have dy/dtheta, but that might come at a significant cost. The derivative of dL/dtheta can be solved by an adjoint problem, and in this case we do not need the complete full matrix. That is particularly relevant for problems with many parameters. However, I think we should move this discussion to a separate issue and consider the current implementation as a starting point for probeye 0.1 (or 1.0 - whatever we think isapprorpiate).

probeye/inference_problem.py
```diff
@@ -562,16 +576,16 @@ def add_experiment(self, exp_name, sensors=None, fwd_model_name=None):
...
-        sensors : dict
+        sensor_values : dict
             The keys are the sensor's names, the values are the measured values.
         fwd_model_name : string
             Name of the forward model this experiment refers to.
         """
 
         # check all keyword arguments are given
-        if sensors is None:
+        if sensor_values is None:
```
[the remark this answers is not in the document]
aklawonn: I wanted to make the user use keyword arguments. But I guess the arguments are very different here (no risk of confusing them) so I converted them into positional arguments. Also I saw (which I did not know before), that you can still provide positional arguments in form of keyword arguments when calling the function, so that's nice.

tests/integration_tests/test_prior_calibration.py
```diff
@@ -0,0 +1,165 @@
...
+
+        # this makes the prior-parameter 'loc_a' a calibration parameter
+        problem.change_parameter_role(
+            'loc_a', new_tex=r"$\mu_a^\mathrm{prior}$",
+            new_info="Std. deviation of normal prior for 'a'",
+            prior=('uniform', {'low': low_loc_a, 'high': high_loc_a}))
+
+        # add the forward model to the problem
+        inp_1 = Sensor("x")
+        out_1 = Sensor("y")
```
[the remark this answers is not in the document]
aklawonn: I changed the code in the forward model's call-function, so that also the y-sensor is used.

tests/integration_tests/test_two_models.py
```diff
@@ -0,0 +1,192 @@
...
+        # ==================================================================== #
+        #                      Define the Forward Models                       #
+        # ==================================================================== #
+
+        class LinearModel(ModelTemplate):
+            def __call__(self, inp):
+                x = inp['x']
+                a = inp['a']
+                b = inp['b']
+                return {'y': a * x + b}
```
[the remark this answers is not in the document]
aklawonn: see above

tests/integration_tests/test_linear_regression.py
```diff
@@ -0,0 +1,178 @@
...
+        # LinearModel above); note that the use of the local parameter name 'm'
+        # for the global parameter 'a' is added here only to highlight the
+        # possibility of this feature; it is not necessary at all here; whenever
+        # forward model's parameter has a similar local and global name (which
+        # should be the case most of the times), one doesn't have to use the
+        # verbose notation  {<global parameter name>: <local parameter name>}
+        # but can instead just write the parameter's (global=local) name, like
+        # it is done with the forward model's parameter 'b' below
+        inp_1 = Sensor("x")
+        out_1 = Sensor("y")
```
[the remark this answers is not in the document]
aklawonn: I see it as Thomas does.  You have both options to define the input. Via an input sensor, as it is currently demonstrated in the linear_regression example, or via Sensor-attributes, as demonstrated in the multiple_sensors example. In fact, you don't have to provided any input sensors at all when defining the forward model (i.e. currently, you have pass an empty list for the input sensors).

examples/example_multiple_sensors.py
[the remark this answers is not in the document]
aklawonn: A sensor has two tasks. First, it connects the model response to the corresponding experimental data by virtue of his name attribute (i.e. the model produces an 'y' value, so there needs to be an 'y' value in the experimental data) which is required for the noise model to work. Second, it allows to define sensor-constants like its position. The latter, however, is essentially also experimental data (e.g. someone measured the sensor position) which should only be stored at one location, which currently is the sensor-object. This sensor-object is currently attached to the forward model, when the latter is instantiated. I would not pass additional instances of the same sensor object (including the constants) to another module because then you have the same data at multiple locations. If these modules need the sensor-constants (e.g. maybe the noise model needs them for some position-correlation), they can just get it from the forward model by referring to the sensor's name. Bottom line is, I would not pass sensor-objects to every module instead of just the sensor's name. Also I would not everywhere pass just the sensor name, because then you can't define sensor-constants.

probeye/definition/forward_model.py
aklawonn: Hey Atul, thanks for your feedback! Currently, I have removed the computation of the jacobian from the code, as it was more of a default for which there will be faster versions in the different inference engines. So if you use Torch, you can just use their methods for computing derivatives without having to consider some existing methods in probeye for that.

probeye/inference/taralli_/postprocessing.py
aklawonn: Good points! Yes, the post-processing will need some further work in the future. Currently, there is just a simple skeleton so that you get some output after calling the taralli routines.

probeye/inference/taralli_/priors.py
aklawonn: Yes indeed, you can define your own priors for Torch. Just check out probeye/inference/taralli_/priors.py to see the priors that are currently used for the taralli setup. In a similar structure, you would write your Torch priors to probeye/inference/pytorch_/priors.py (not existing yet).

probeye/definition/noise_model.py
```diff
@@ -0,0 +1,154 @@
...
+        self.experiment_names = []
+
+    def add_experiment_names(self, experiment_names_, problem_experiments):
+        """
+        Adds experiment names to the noise model. When the noise model is
+        evaluated it will only be evaluated for those experiments added here.
+
+        Parameters
+        ----------
+        experiment_names_ : str, list[str]
```
joergfunger, line 40: not sure what the difference is here, one is the name, the other the actual data structure? Why do I need both?

probeye/definition/forward_model.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class ForwardModelTemplate:
+    """
+    This class serves as a base class for any forward model. When you want to
+    define a specific forward model, you need to derive your own class from this
+    one, and then define the '__call__' method. The latter essentially describes
+    the model function mapping the model input to the output.
+    """
+    def __init__(self, prms_def_, input_sensors, output_sensors):
```
joergfunger, line 14: I'm still struggling with the difference between prms and input_sensors, are the first ones the variable ones to be calibrated and the second one are the measured deterministic data? And why is that important for someone writing the forwardmodel template?

probeye/definition/noise_model.py
```diff
@@ -0,0 +1,154 @@
+# third party imports
+import numpy as np
+
+# local imports
+from probeye.subroutines import len_or_one
+from probeye.subroutines import list2dict, make_list
+
+
+class NoiseModelTemplate:
```
joergfunger, line 9: Is that the Baseclass, or a template that others should copy from? Isn't this class rather a special uncorrelated Gaussian noise class instead of a template?

probeye/definition/sensor.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        Parameters
+        ----------
+        name : str
+            The name of the sensor, e.g. 'Deflection-Sensor bottom-left'.
+        """
+        self.name = name
+
+
+class PositionSensor(Sensor):
```
joergfunger, line 18: I would provide the coordinates in a single array (makes using that easier, because you can directly access all coordinates by extracing a single row. The dimension would IMO also be flexible.

joergfunger (commented):
There are only a couple of small remarks left. The only remaining thing is the separation between pms and input sensors, which I don't think is intuitive.

probeye/definition/noise_model.py
aklawonn: This is a template-class, other noise models should be derived from. An example is the noise class in probeye/inference/taralli_/noise_models.py. I don't see why it would be a special case of an uncorrelated normal noise, since it does not compute anything.

probeye/definition/noise_model.py
aklawonn: The first argument states a list of experiment-names that are in the scope of the noise model. The second argument is essentially the experiment database of the problem (the dict with all the experimental data). This second argument is only used to check if the experiment names given in the first argument appear in the problem's experimental database, and if those experiments are consistent with the noise model (i.e. if they have all sensors that the noise model requires). Long story short: the second argument is just for checking the first argument.

probeye/definition/noise_model.py
joergfunger: What about the error, isn't that already some computation?

probeye/definition/sensor.py
aklawonn: I added an attribute with all coordinates as a vector. However, I would leave the dimension at three for this simple PositionSensor. A sensor class with more than three coordinates sounds like a custom position sensor class to me.

probeye/definition/noise_model.py
aklawonn: No, since it requires the error_function which is undefined in this class.

joergfunger: I think the question is, if all noise models need an error function that is called by some tool - if this is not the case then this would not necessarily be a method of the base class?

probeye/definition/forward_model.py
aklawonn: Input sensors contain measured data that - in general - differs from experiment to experiment, and which is required by the forward model as an input argument. Those values are always fixed, deterministic and will not be inferred, which distinguishes them from the prms-argument, which are the parameters that should be inferred.

probeye/definition/noise_model.py
aklawonn: I would think that this setup is useful for the majority of the problems. However, this does not mean that it will always apply. There are probably cases where you need something different. But then, you would just overwrite the error method with your custom error-method. I think for most users it will be helpful to see how this default method looks like compared to the case where we would provide no guidance at all.

joergfunger: For me, the noise model is not yet a pure problem description, or vice versa not sufficient to describe all possible noise models (independent of the implementation), e.g. Gaussian normal noise is a standard that should be part of the problem definition library (not the inference engine).

probeye/definition/noise_model.py
aklawonn: The second argument is now an attribute of the noise model, and hence not required as an input anymore.

probeye/definition/sensor.py
joergfunger: what about dimensions 1 and 2?

aklawonn: I updated the PositionSensor for the 1D and 2D case. It's already pushed.

probeye/definition/noise_model.py
aklawonn: As discussed, I moved the NormalNoise to the definition part of the package.

aklawonn: Also, all Template classes were renamed to Base classes.

probeye/inference/taralli_/postprocessing.py
aklawonn: This discussion is continued in issue #8.

probeye/definition/forward_model.py
aklawonn: As discussed on the phone, the name input sensors will be kept for now.

examples/example_linear_regression.py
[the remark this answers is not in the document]
aklawonn: As discussed on the phone, the sensor terminology will be kept for now.

examples/example_multiple_sensors.py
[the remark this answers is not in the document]
aklawonn: As discussed on the phone, the sensor terminology will be kept for now.

examples/example_multiple_sensors.py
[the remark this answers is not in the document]
aklawonn: As discussed on the phone, the sensor terminology will be kept for now.

probeye/forward_model.py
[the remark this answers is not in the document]
aklawonn: As discussed on the phone, the sensor terminology will be kept for now.

probeye/forward_model.py
aklawonn: As discussed on the phone, the sensor terminology will be kept for now.

probeye/forward_model.py
aklawonn: The error functions - as discussed - have been moved to the noise model.

tests/integration_tests/test_linear_regression.py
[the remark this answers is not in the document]
aklawonn: As discussed on the phone, the sensor terminology will be kept for now.

probeye/inference_problem.py
aklawonn: I opened an issue for the correlation methods (see issue #9).

probeye/inference_problem.py
aklawonn: As discussed, calibration-parameters are now called latent parameters.

probeye/inference_problem.py
aklawonn: As discussed, calibration-parameters are now called latent parameters.

probeye/noise.py
aklawonn: By now, one can simply choose between an additive or a multiplicative error function.

probeye/definition/inference_problem.py
```diff
@@ -0,0 +1,1075 @@
...
+
+    @property
+    def noise_prms(self):
+        """Provides noise_prms attribute."""
+        return self._parameters.noise_prms
+
+    @property
+    def parameters(self):
+        """Access self._parameters from outside via self.parameters."""
+        return self._parameters
```
TTitscher, line 141: What is the advantage of this decorator over just storing `self.parameters`?

aklawonn: These private attributes should not be directly edited from outside, but they should be accessible. So, by the current setup you can access the parameters attribute via problem.parameters, but you cannot edit it e.g. via problem.parameters['a'] = {...}.

TTitscher: I agree with the behavior you describe, but the code does not complain when I do
`problem.parameters["a"].type = 5`
or 
`problem.parameters["new_prm"] = ParameterProperties({...})`

aklawonn: Good point, I opened an issue on that matter (issue #10). However, I did not do anything about it yet as this does not affect the normal use of probeye.

joergfunger (approved):

PR #13, opened 2021-09-19
The attributes of the InferenceProblem's parameters cannot be changed…

### LINKED ISSUE #10: Better protection for private attributes of InferenceProblem
Currently, you can modify the private attributes of InferenceProblem (such as InferenceProblem._parameters) via the corresponding properties. This means, for example, you modify InferenceProblem._parameters['a'] via
`InferenceProbelm.parameters['a'] = InferenceProbelm.parameters['b']`
This should not be possible in order to prevent messing with private attributes. Preferable, there should be an error raised, if something like this is tried.

aklawonn:
… directly anymore. This fixes #10

### DIFF
--- a/probeye/definition/inference_problem.py
+++ b/probeye/definition/inference_problem.py
@@ -401,12 +401,25 @@ class InferenceProblem:
                 self.remove_parameter(prior_prm)  # recursive call
             del self._priors[self._parameters[prm_name].prior.name]
             del self._parameters[prm_name]
-            # correct the indices of the remaining 'latent'-parameters
+            # correct the indices of the remaining 'latent'-parameters; note
+            # that the way how the correction is done is due to the fact that
+            # the parameter.index attribute is protected, and cannot be changed
+            # directly from outside
+            idx_dict = {}
             idx = 0
-            for name, parameter in self._parameters.items():
+            for prm_name, parameter in self._parameters.items():
                 if parameter.index is not None:
-                    parameter.index = idx
+                    idx_dict[prm_name] = idx
                     idx += 1
+            for prm_name, idx in idx_dict.items():
+                self._parameters[prm_name] = ParameterProperties(
+                    {'index': idx,
+                     'type':  self._parameters[prm_name].type,
+                     'role':  self._parameters[prm_name].role,
+                     'prior': self._parameters[prm_name].prior,
+                     'value': self._parameters[prm_name].value,
+                     'info':  self._parameters[prm_name].info,
+                     'tex':   self._parameters[prm_name].tex})
 
     def check_if_parameter_exists(self, prm_name):
         """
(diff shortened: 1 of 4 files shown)


probeye/definition/inference_problem.py
```diff
@@ -494,26 +507,37 @@ def change_parameter_role(self, prm_name, const=None, prior=None,
...
         """
         # check if the given parameter exists
         self.check_if_parameter_exists(prm_name)
 
+        # if None is given for the new info/tex, the old value will be kept
+        if new_info is None:
+            new_info = self._parameters[prm_name].info
+        if new_tex is None:
+            new_tex = self._parameters[prm_name].tex
+
```
TTitscher, line 531: How about
~~~py
self._parameters[prm_name]._info = new_info
self._parameters[prm_name]._tex = new_tex
~~~
instead of creating a new object with almost identical attributes?

aklawonn: Good point, that's what I tried at first. But then pycharm is not happy because you shouldn't modify the private attributes of ParameterProperties-object from outside.

TTitscher (commented):
As discussed in the related issue, I have my doubts here. The whole issue with setters and getters somehow _feels_ not pythonic.
* Read access to the parameter attributes is (only) possible via the attributes, write access is handled indirectly via the inference problem. For me, that is inconsistent.
    * If the design goal is to access and modify all information via the inference problem, then there should be `get_parameter_info(...)`, `get_parameter_tex(...)` and so on. And at this point, I would question this design...
* I see no harm in changing the parameter `info` and `tex` directly and save 50 or so LOC. 

Instead of writing all that code to prevent a user from from using the parameters class directly, I'd rather solve the underlying issue on _why_ this is now allowed. I'll have a closer look, but I assume it is related to having "role" and "prior" at the parameter. 

Maybe we need another opinion here...

aklawonn:
First of all, thank you for taking the time to check out my modifications so quickly! So, let me explain. The main idea of protecting the attributes is that (in general) you cannot change one of them without having to change something else as well if you don't want to mess up your problem. For example, you cannot just change a latent parameter's role to 'const', without taking care of the prior and the prior's parameters that were attached to this latent parameter. I don't see this as an underlying issue. Of course we could let the user do what she wants, also if it messes up the problem, but for me this approach is not attractive if this tool should not only appeal to a handful of experts. However, I agree that in this case, the info and tex attribute don't need to be protected. I changed that in the last commit.

aklawonn:
One more comment: you write that you find it inconsistent that "Read access to the parameter attributes is (only) possible via the attributes, write access is handled indirectly via the inference problem". For me, this is not inconsistent, but exactly the goal of this issue. The write access is handled by the inference problem (are permitted entirely), because it also takes care all cleanup-tasks that might be necessary, when you want to change one of the protected parameter properties. Read access however, is unrestricted, because there is no harm in reading these properties.

aklawonn:
As discussed on the phone, I will merge this branch as it is now, and @TTitscher will create a new issue regarding a decoupling ("ent-wurbeling") of dependencies in the problem's parameter structure.

PR #23, opened 2021-09-23
Coupling parameters and priors

### LINKED ISSUE #21: Coupling between parameters and priors
We currently store information related to the prior in three places
1. `inference_problem.prior` (a dict with key `prm_name`)
2. `inference_problem.parameters["..."].prior `
3. `inference_problem.parameters["..."].role` (saying "latent" or "const")

This duplicate information makes it easy for the user to create inconsistent states and we need to restrict the access to the objects (#13) and need to implement methods like `change_parameter_role` and `change_constant` to keep everything in sync.

Suggestion:
* Assume: `inference_problem.prior` is the only place to store prior information and a parameter "x" is latent, `if "x" in inference_problem.priors`. 
* remove attributes `latent` and `prior` from the `ParameterProperties`

Then, it is much harder to create invalid states and we could simplify the related methods, e.g.
* `change_parameter_role` could become adding/removing its prior from `problem.priors`
* `change_constant` could become `problem.parameters["my_prm"].value = 42.`
* ...

aklawonn:
This fixes #21

### DIFF
--- a/probeye/definition/inference_problem.py
+++ b/probeye/definition/inference_problem.py
@@ -425,7 +425,6 @@ class InferenceProblem:
         prm_name : str
             A global parameter name.
         """
-        # check if the given parameter exists
         if prm_name not in self._parameters.keys():
             raise RuntimeError(
                 f"A parameter with name '{prm_name}' has not been defined yet.")
@@ -458,7 +457,6 @@ class InferenceProblem:
         new_tex : str or None, optional
             The new string for the parameter's tex-representation.
         """
-        # check if the given parameter exists
         self.check_if_parameter_exists(prm_name)
 
         # exactly one of the const and prior key word arguments must be given
@@ -514,7 +512,6 @@ class InferenceProblem:
         new_tex : str, None, optional
             The new string for the parameter's tex-representation.
         """
-        # check if the given parameter exists
         self.check_if_parameter_exists(prm_name)
 
         # if None is given for the new info/tex, the old value will be kept
@@ -544,7 +541,6 @@ class InferenceProblem:
         new_value : float
             The new value that prm_name should assume.
         """
-        # check if the given parameter exists
         self.check_if_parameter_exists(prm_name)
 
         # check if the given parameter is a constant
--- a/probeye/definition/inference_problem.py
+++ b/probeye/definition/inference_problem.py
@@ -429,8 +429,7 @@ class InferenceProblem:
             raise RuntimeError(
                 f"A parameter with name '{prm_name}' has not been defined yet.")
 
-    def change_parameter_role(self, prm_name, const=None, prior=None,
-                              new_info=None, new_tex=None):
+    def change_parameter_role(self, prm_name, const=None, prior=None):
         """
         Performs the necessary tasks to change a parameter's role in the problem
         definition. A parameter's role can either be changed from 'const' to
@@ -452,10 +451,6 @@ class InferenceProblem:
             parameter within the problem scope. An example for a normal prior:
             ('normal', {'loc': 0.0, 'scale': 1.0}). In order to define the
             prior's parameters, check out the prior definitions in priors.py.
-        new_info : str or None, optional
-            The new string for the explanation of parameter prm_name.
-        new_tex : str or None, optional
-            The new string for the parameter's tex-representation.
         """
         self.check_if_parameter_exists(prm_name)
 
@@ -484,20 +479,10 @@ class InferenceProblem:
         # the parameter's role is changed by first removing it from the problem,
         # and then adding it again in its new role; the role-change does not
         # impact the type ('model', 'prior' or 'noise')
-        prm_type = self._parameters[prm_name].type
-        # if no new_info/new_tex was specified, use the old ones
-        if new_info is None:
-            prm_info = self._parameters[prm_name].info
-        else:
-            prm_info = new_info
-        if new_tex is None:
-            prm_tex = self._parameters[prm_name].tex
-        else:
-            prm_tex = new_tex
-        # now we can finally change the role
+        prm = self._parameters[prm_name]
         self.remove_parameter(prm_name)
-        self.add_parameter(prm_name, prm_type, const=const, prior=prior,
-                           info=prm_info, tex=prm_tex)
+        self.add_parameter(prm_name, prm.type, const=const, prior=prior,
+                           info=prm.info, tex=prm.tex)
 
     def change_parameter_info(self, prm_name, new_info=None, new_tex=None):
         """
--- a/probeye/definition/inference_problem.py
+++ b/probeye/definition/inference_problem.py
@@ -7,8 +7,7 @@ from tabulate import tabulate
 import numpy as np
 
 # local imports
-from probeye.definition.parameter import Parameters, ParameterProperties
-from probeye.definition.prior import PriorBase
+from probeye.definition.parameter import Parameters
 from probeye.subroutines import underlined_string, titled_table
 from probeye.subroutines import simplified_list_string, simplified_dict_string
 from probeye.subroutines import unvectorize_dict_values, make_list, len_or_one
--- a/probeye/definition/inference_problem.py
+++ b/probeye/definition/inference_problem.py
@@ -408,7 +408,7 @@ class InferenceProblem:
                     idx += 1
             for prm_name, idx in idx_dict.items():
                 self._parameters[prm_name] =\
-                    self._parameters[prm_name].return_changed_copy(index=idx)
+                    self._parameters[prm_name].changed_copy(index=idx)
 
     def check_if_parameter_exists(self, prm_name):
         """
@@ -502,7 +502,7 @@ class InferenceProblem:
 
         # change the info/tex-string
         self._parameters[prm_name] =\
-            self._parameters[prm_name].return_changed_copy(info=new_info,
+            self._parameters[prm_name].changed_copy(info=new_info,
                                                            tex=new_tex)
 
     def change_constant(self, prm_name, new_value):
@@ -526,7 +526,7 @@ class InferenceProblem:
             )
         # change the parameter's value
         self._parameters[prm_name] =\
-            self._parameters[prm_name].return_changed_copy(value=new_value)
+            self._parameters[prm_name].changed_copy(value=new_value)
 
     def check_problem_consistency(self):
         """
--- a/probeye/definition/inference_problem.py
+++ b/probeye/definition/inference_problem.py
@@ -1044,16 +1044,13 @@ class InferenceProblem:
 
         # check that each globally defined experiment appears in one of the
         # noise models
+        exp_names_in_noise_models = set()
+        for noise_model in self._noise_models:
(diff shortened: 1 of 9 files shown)


probeye/definition/inference_problem.py
```diff
@@ -368,7 +365,6 @@ def add_parameter(self, prm_name, prm_type, const=None, prior=None,
         # add the parameter to the central parameter dictionary
         self._parameters[prm_name] = ParameterProperties({'index': prm_index,
```
TTitscher, line 366: When I look at this method, it basically _creates_ a `ParameterProperties` object and assigns it to `self._parameters`. So I would suggest putting all that code with the checks and prior creation into `ParameterProperties.__init__` and 
have implement the `add_parameter` method here as a simple forward
~~~py
def add_parameter(name, type, ...):
    self.parameters[name] = ParameterProperties(type, ...)
~~~

I personally would remove `InferenceProblem.add_parameter`, so replace
~~~py
problem.add_parameter("p", "model", const=42) 
~~~
by
~~~py
problem.parameters["p"] = ParameterProperties("model", const=42)
~~~
But I suspect you dislike that :smile:

probeye/definition/parameter.py
```diff
@@ -183,14 +184,100 @@ def __init__(self, prm_dict):
...
-        self._role = prm_dict['role']
         self._prior = prm_dict['prior']
         self._value = prm_dict['value']
         self.info = prm_dict['info']
         self.tex = prm_dict['tex']
 
+        # check the given values
+        self.check_consistency()
+
+    def check_consistency(self):
```
TTitscher, line 198: I thought the whole point of the methods `add_parameter`, `change_role` or `remove_parameter` is to change the parameters in a consistent way. And those methods itself are full of checks. 
That is why I am surprised to find more checks here.

probeye/definition/parameter.py
```diff
@@ -215,8 +302,9 @@ def type(self, value):
 
     @property
     def role(self):
```
TTitscher, line 304: If there is only one of two options, "latent" or "const" (aka "not_latent"), I'd prefer a bool like
~~~py
def is_latent(self):
   return self._index is not None
~~~

probeye/definition/parameter.py
```diff
@@ -183,14 +184,100 @@ def __init__(self, prm_dict):
...
+                f"Found invalid ParameterProperties._index attribute! It must "
+                f"be of type int or None, but found {type(self._index)}.")
+
+        if (self._index is not None) and (self._index < 0):
+            raise RuntimeError(
+                f"Found negative value for ParameterProperties._index! This "
+                f"attribute must be a non-negative integer, but found a value "
+                f"of {self._index}.")
+
+        if type(self._type) != str:
```
TTitscher, line 220: I think this is pretty much covered in the check below of `self._type not in ["model", ...]`

probeye/definition/parameter.py
```diff
@@ -183,14 +184,100 @@ def __init__(self, prm_dict):
...
+            raise TypeError(
+                f"Found invalid ParameterProperties._value attribute! It must "
+                f"be of type float/int or None, but found {type(self._value)}.")
+
+        if type(self.info) != str:
+            raise TypeError(
+                f"Found invalid ParameterProperties.info attribute! Its type "
+                f"must be str, but found {type(self.info)}.")
+
+        if not (type(self.tex) == str or self.tex is None):
```
TTitscher, line 246: Why is `tex` and `info` required to be a string? Does the library rely on that? The essence of python duck typing is that you actually do not check the types but their behavior.

Actually, I could think of a `TeXLabel` class with some fancy `__str__` method that is able to change the format. As it has a `__str__` method (like every class), I see no reason to forbid that.

TTitscher (approved):
Exaggeratedly said, the code is 75% checks and runtime errors. The checks _after_ object creation as well as some type checks are IMO unnecessary.

probeye/definition/inference_problem.py
aklawonn: How about this: I try to put the checks and most of the code of add_parameter into the Parameters and ParameterProperties classes, but we keep the add_parameter method. As you suspected, I indeed prefer such a method over a direct statement that you prefer ;-) I like the add_parameter method because is more verbose (it says what is happening), and does not require any knowledge about the structure of how the parameters are stored.

probeye/definition/parameter.py
aklawonn: You're right, there's a lot of checks :D These additional checks might be overkill - I added them, because I don't know if there really is no way to mess things up with the parameters.

probeye/definition/parameter.py
aklawonn: That's funny, in a previous draft, I had changed the role attribute to exactly the attribute you propose. I changed it back essentially because this attribute does not say what it is if it is not latent (is_latent=False => okay, it's not latent... but what is it then?). But I don't mind going back to the is_latent attribute.

probeye/definition/parameter.py
aklawonn: You're right, I'll remove it.

probeye/definition/parameter.py
aklawonn: Sure, I would not mind that at all :) However, IMO this is really a thing in the category 'nice to have', but not essential. But if you want to implement that, just go ahead.

TTitscher: I do not propose to implement a `TexLabel` it the library. I wanted to point out that if I (as a user) implement that in my code, I would be not allowed to use it, as it is not of type string.

probeye/definition/parameter.py
TTitscher: The current code says that the role is either "latent" or "const". By that, I'd say "is_latent=False" means a constant parameter that is not inferred.

aklawonn: Yes, sure, I was looking on it from the perspective of someone, who does not know what other options there are.

probeye/definition/parameter.py
aklawonn: Okay, if you want to keep such a door open, I'll remove these explicit checks from the code.

aklawonn:
This is just a detail, but I think code is more readable when functions/methods describe what they do. This is why I changed the 'changed'-method in ParameterProperties to 'change'.

TTitscher:
>when functions/methods describe what they do

Agreed! But the whole point is that this method _does not_ change the object but returns a changed copy. So how about `changed_copy`?

probeye/definition/parameter.py
aklawonn: I just removed it.

aklawonn:
> Agreed! But the whole point is that this method does not change the object but returns a changed copy.
Okay, technically, this is of course correct, although I would have seen that as the way how the instance is changed (even if technically it is not actually changed itself). I don't like the name _changed_copy_ so much, because, again, it does not describe what it does. How about _return_changed_copy_ or _return_modified_copy_, even though I see that these names are a bit bulky. Btw, I guess now we have reached the current top of detail-discussing xD.

TTitscher:
Please not "return_..." :confused:  . I mean likewise, you would call the [`loglike_contribution`](https://github.com/BAMresearch/probeye/blob/main/probeye/definition/noise_model.py) method `compute_and_return_loglike_contribution` or so, because 
~~~
log_like += noise.loglike_contribution(...)
~~~
reads just fine. Even though it _does_ compute and return it...

aklawonn:
Still working on it...

probeye/definition/parameter.py
```diff
@@ -318,6 +389,18 @@ def role(self, value):
             "You cannot change a parameter's role directly! Use "
             "InferenceProblem.change_parameter_role instead.")
 
+    @property
+    def is_latent(self):
+        """Adds a pseudo-attribute self.is_latent, which allows a convenient
+           check on whether a parameter is latent or not."""
+        return True if self._index is not None else False
```
TTitscher, line 396: How about
~~~py
return self._index is not None
~~~

probeye/definition/parameter.py
```diff
@@ -318,6 +389,18 @@ def role(self, value):
...
+
+    @property
+    def is_const(self):
+        """Adds a pseudo-attribute self.is_const, which allows a convenient
+           check on whether a parameter is constant or not."""
+        return True if self._index is None else False
```
TTitscher, line 402: You currently base the decision on latent/const on `self._index`. If you change that, you need to modify both the `is_latent` and `is_const` method. You could avoid that (e.g.) by
~~~py
return not self.is_latent
~~~

probeye/definition/parameter.py
```diff
@@ -24,6 +24,77 @@ def __setitem__(self, key, value):
...
+        Checks if a parameter, given by its name, exists among the currently
+        defined parameters. An error is raised when the given parameter does
+        already exist.
+
+        Parameters
+        ----------
+        prm_name : str
+            A global parameter name.
+        """
+        if prm_name in [*self.keys()]:
```
TTitscher, line 94:
~~~py
if prm_name in self:
~~~
would be the common way to check that.

probeye/definition/inference_problem.py
```diff
@@ -447,7 +397,8 @@ def change_parameter_role(self, prm_name, const=None, prior=None):
             ('normal', {'loc': 0.0, 'scale': 1.0}). In order to define the
             prior's parameters, check out the prior definitions in priors.py.
         """
-        self.check_if_parameter_exists(prm_name)
+        # first, make sure that the given parameter exists
+        self._parameters.confirm_that_parameter_exists(prm_name)
```
TTitscher, line 401: The alternative would be a simple
~~~py
assert prm_name in self._parameters
~~~
which is IMO more expressive and the resulting error message has all the information in it, I'd say. Just an idea.

probeye/definition/parameter.py
aklawonn: good point, done!

probeye/definition/parameter.py
aklawonn: I'll change that

probeye/definition/parameter.py
aklawonn: indeed, will be changed

probeye/definition/inference_problem.py
aklawonn: Now I moved almost all of the code of add_parameter/remove_parameter to the Parameters-class.

probeye/definition/parameter.py
aklawonn: Most of the checks are now within the Parameters-class. This should have eliminated several redundant checks.

probeye/definition/parameter.py
aklawonn: I added a 'is_latent' and a 'is_constant' attribute to the ParameterProperties-class.

probeye/definition/inference_problem.py
aklawonn: I can see that this is nice and short, but I like the verbosity and the custom error message of the current implementation. If you don't mind, I would like to keep that.

PR #24, opened 2021-09-27
Parameter properties changed

TTitscher:
Just a brief idea on how to keep the ParameterProperties immutable but also simplify their creation.


aklawonn:
Nice! Maybe you can additionally provide some explanation (docstring) on the "changed" method, so that one quickly understands what the method is for.

aklawonn (approved):

PR #39, opened 2021-11-14
Improve dependency management

rozsasarpi:
## Why

The current dependency specification (`install_requires` in `setup.cfg`) allows to install the latest version of each package that `probeye` is dependent on. This means that if one of the packages increases its main version number that version will be installed. However, if we assume that the package developers follow semantic versioning that means that at least one breaking change was introduced with the major version number increase, which in turn may break `probeye`.

## What

* Major version number upper limits are added to the `install_requires` dependencies.
* It is safer (causes fewer errors in CI pipelines and on the users' side) to disallow major version number increase of dependencies compared to their tried/tested highest version number.
* Based on my experience with python packages it is ok to assume that package developers follow semantic versioning (although there are many exceptions).
* If a dependency increases its major version number that can be tried/tested locally and/or on a separate branch and then the after taking the necessary actions (often no action is required) the major version number limit can be increased manually in the `setup.cfg` file.

## Testing

* GitHub actions
* local install

## Other

* I used semantic commit messages, following the guidelines outlined here: https://www.conventionalcommits.org/en/v1.0.0/. @aklawonn if you prefer to follow another convention please let me know.
* I recommend to squash the commits in the pull request (I think it is not yet allowed in the repo).

### DIFF
--- a/.github/workflows/push.yaml
+++ b/.github/workflows/push.yaml
@@ -5,6 +5,8 @@ name: CI
 
 on:
   push:
+  schedule:
+    - cron: '0 1 * * 6'  # weekly run at 01:00 UTC on Saturday, dependency check
 
 jobs:
   # Lint and type checking
@@ -82,6 +84,37 @@ jobs:
           label: coverage
           message: ${{ env.COVERAGE_PCT }}
           color: ${{ env.COVERAGE_COL }}
+
+  # Run tests with the latest available version of all dependencies
+  # these latest versions might be beyond the versions specified in `setup.cfg`
+  # the goal of this job to catch errors with these latest versions
+  # [only for `main`]. Ideally, if this job succeeded with at least one major
+  # version number increased then the `setup.cfg` would be updated.
+  test_with_latest:
+      name: Test with latest dependencies on Python ${{ matrix.python-version }}
+      runs-on: ubuntu-latest
+      if: |
+        github.ref == 'refs/heads/main' &&
+        github.repository == 'BAMresearch/probeye' &&
+        github.event_name == 'schedule'
+      needs: [ lint_and_type_check ]
+      strategy:
+        matrix:
+          python-version: [ 3.6, 3.7, 3.8, 3.9 ]
+
+      steps:
+        - uses: actions/checkout@v2
+        - name: Set up Python ${{ matrix.python-version }}
+          uses: actions/setup-python@v2
+          with:
+            python-version: ${{ matrix.python-version }}
+        - name: Install package with latest dependencies
+          run: |
+            python probeye/_setup_cfg.py
+            pip install .[tests]
+        - name: Test with pytest
+          run: |
+            pytest
 
   # Release to GitHub and PyPi [only for `main`]
   #release:
--- a/probeye/_setup_cfg.py
+++ b/probeye/_setup_cfg.py
@@ -1,0 +1,74 @@
+"""This module contains functions that remove version number constraints from
+selected packages in the setup.cfg file. It is meant to be used during CI: to test if
+the package works with the latest versions of its dependencies."""
+
+import configparser
+import re
+from typing import List
+
+
+def version_constraint_free_packages(setup_cfg_packages: str) -> List[str]:
+    """
+    Remove version constraints from the packages listed in `options_field` (provided
+    as input argument or read from `setup.cfg`).
+
+    Parameters
+    ----------
+    setup_cfg_packages: str
+        Packages with optional version constraints as parsed from a syntactically
+        correct `setup.cfg` file.
+
+    Returns
+    -------
+    List[str]
+        `install_requires` packages without version constraints.
+    """
+    # to be able to match the last package in all cases
+    setup_cfg_packages = "\n" + setup_cfg_packages + "\n"
+
+    # get plain (version constraint free) package names, they are listed using a
+    # "list-semi": dangling list or string of semicolon-separated values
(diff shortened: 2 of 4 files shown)


joergfunger (commented):
Wouldn't it be nice to somehow automatically test this. In the proposed setup, we won't even realize that a package was updated and that a newer version fails.

aklawonn:
I think that's a good idea, Arpad! On your testing proposal, Jörg, I think it might need some non-negligible overhead for an event that does not happen very often. Alternatively, we would just check from time to time, if there are newer major versions, and test locally, if those also work (and then adjust setup.cfg if there are not problems). And even if there were newer major versions, that we haven't noticed yet, it would not really be a problem for us if we use an older major version that works with probeye, right?

joergfunger:
Well, I think think from experience that check will not happen very often, and thus I think it would be good to have two separate versions (one release with the versions set as Arpad proposed, but another one with the current versions) being tested. Alternatively, we could just have a script integrated into the CI that automatically checks for updates and provides that information (without actually testing our implementation).

aklawonn:
From those two, I'd prefer the second option. In the first one, we'd have twice the testing time, while most of the times the identical tests will be run (when no updates happend).

rozsasarpi:
I agree that it would be nice to regularly check the compatibility with the latest versions of the dependencies. (Ideally the compatible version numbers would be even updated automatically if all tests are passed.)

> In the first one, we'd have twice the testing time, while most of the times the identical tests will be run twice (when no updates happend).

With the generous GitHub action support of Microsoft for public GitHub repos, I would not be concerned with this.

I usually have a weekly scheduled CI pipeline to catch if a non-major version number upgrade breaks the code (to cover the packages that do not follow semantic versioning). The same scheduled CI pipeline could have a run where the dependency version numbers are not restricted. I can look into how a second dependency list could be set up in a convenient manner.

rozsasarpi:
> In the proposed setup, we won't even realize that a package was updated and that a newer version fails.

> it would be good to have two separate versions (one release with the versions set as Arpad proposed, but another one with the current versions) being tested.

I could not find an established/reliable workflow or tool to do this. Surprisingly, `pip`'s dependency resolver is not complete so "update all" at the moment is not available, see https://github.com/pypa/pip/issues/4551. The issue has been open for years.

To still be able to handle our particular case and to implement @joergfunger's suggestion, I propose the following setup:

* run a scheduled CI pipeline using GitHub Actions (e.g. once a week) [code](https://github.com/rozsasarpi/probeye/blob/d55882bf45aab9bf04d58cea0972bb812b70a33a/.github/workflows/push.yaml#L9)
* only for the scheduled pipeline, run all the tests with a `probeye` installation that uses the latest available versions of its dependencies (no version limit constraint) [code](https://github.com/rozsasarpi/probeye/blob/d55882bf45aab9bf04d58cea0972bb812b70a33a/.github/workflows/push.yaml#L55-L79)
* [this line](https://github.com/rozsasarpi/probeye/blob/d55882bf45aab9bf04d58cea0972bb812b70a33a/.github/workflows/push.yaml#L75) does the trick, i.e. overwriting the `setup.cfg` file with a version that differs from the original one only by the lack of version number constraints for packages listed in `install_requires`.
*  [this is the function](https://github.com/rozsasarpi/probeye/blob/d55882bf45aab9bf04d58cea0972bb812b70a33a/probeye/_setup_cfg.py) that does the job. I think function covers all valid syntax of `install_requires`.

Note that this overwrite of the `setup.cfg` is not propagated back to the repo, it is only on the runner.

This setup allows to test `probeye` with the latest version of its dependencies without affecting the users. The developers get an email when the scheduled pipeline fails so they can take action if they wish.

What is still missing: if the scheduled pipeline with the latest dependencies succeeds then we do not get a notification, we still have the unnecessarily restrictive version constraints in the `setup.cfg`. I consider this to be acceptable.

Please let me know what you think. I wanted to hear your thoughts before committing the linked changes to this branch.

aklawonn:
Thanks for looking into this and providing a solution so quickly! For me this is totally sufficient. One question though. You wrote:

>  if the scheduled pipeline with the latest dependencies succeeds then we do not get a notification, we still have the unnecessarily restrictive version constraints in the setup.cfg. 

I assume these workflows would be logged, so that we could check on them online. So our approach would then be to once in a while check on those weekly run-logs, and then - given there were major package updates that did not result in any problems - update the package restrictions in the setup.cfg file by hand. Did I understand that correctly?

rozsasarpi:
Yes, your understanding is correct. The logging is there automatically the same way as for other CI pipelines and you can see all their details, e.g. see them for the forked repo: https://github.com/rozsasarpi/probeye/actions. You can see immediately which ones are the scheduled ones and by clicking on them you can get more detailed information.

image

aklawonn:
Nice! Then from my side, we can merge. But I don't know what Thomas and Jörg think.

joergfunger:
Shouldn't we then add the files mentioned above to the current pull request, or are we opening another one for adding this automatic testing of updated package versions?

rozsasarpi:
I was planning to add the files and changes here if the reviewers agree that it is a good approach.

joergfunger:
From my point of view, this would be great.

rozsasarpi:
@TTitscher what's your take on the proposed changes?

TTitscher (approved):
Sorry, I kinda missed that discussion. I like the concept/idea as well as the implementation in `_setup_cfg.py`. Thanks!

PR #50, opened 2021-11-22
code style

aklawonn:
Let's address the code style/formatting issue. I'm trying to get the black code formatter into the workflow. Also, all files should be reformatted here according to the black format. Finally, let's try out type hinting.

### DIFF
--- a/probeye/definition/forward_model.py
+++ b/probeye/definition/forward_model.py
@@ -1,8 +1,10 @@
 # standard library
+#from typing import Union
 import copy as cp
 import numpy as np
 
 # local imports
+#from probeye.definition.sensor import Sensor
 from probeye.subroutines import translate_prms_def
 from probeye.subroutines import make_list
 from probeye.subroutines import len_or_one
@@ -82,7 +84,7 @@ class ForwardModelBase:
         """Provides input_sensor_names attribute."""
         return [sensor.name for sensor in self.output_sensors]
 
-    def response(self, inp):
+    def response(self, inp: dict) -> dict:
         """
         Evaluates the model response and provides computed results for all of the
         model's output sensors. This method must be overwritten by the user.
@@ -104,13 +106,13 @@ class ForwardModelBase:
             "define this method, so you can evaluate your model."
         )
 
-    def __call__(self, inp):
+    def __call__(self, inp: dict) -> dict:
         """
         Calls the self.response method. Shortens internal forward model calls.
         """
         return self.response(inp)
 
-    def jacobian(self, inp):
+    def jacobian(self, inp: dict) -> dict:
         """
         Numerically computes the Jacobian matrix of the forward model and returns it in
         form of a dictionary. Note that this method should be overwritten, if there is a
@@ -195,10 +197,10 @@ class ForwardModelBase:
 
     def jacobian_dict_to_array(
         self,
-        inp,
-        jac_dict,
-        n_inp_dim,
-    ):
+        inp: dict,
+        jac_dict: dict,
+        n_inp_dim: int,
+    ) -> np.ndarray:
         """
         Converts the Jacobian in dict-format (computed by the above 'jacobian' method)
         into a numpy array. This method is external to the above 'jacobian' method, so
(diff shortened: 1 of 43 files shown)


rozsasarpi:
Thanks for picking this up. I think you do not want to delete the `.pre-commit-config.yaml` file. It describes the checks/rules that you want to pass and it allows to recreate the same environment locally (for those -- like me -- who uses pre-commit hooks) and also during the CI pipeline. So in my opinion you should keep it and extend it with `black`. Something like this:

```
repos:
- repo: https://github.com/pre-commit/pre-commit-hooks
  rev: v4.0.1
  hooks:
    - id: check-added-large-files
      args: ['--maxkb=100']
    - id: check-yaml

- repo: https://github.com/psf/black
  rev: 21.5b2
  hooks:
    - id: black
```

aklawonn:
Not quite sure where this coverage error is coming from...

rozsasarpi:
[This entire section](https://github.com/BAMresearch/probeye/blob/main/.github/workflows/push.yaml#L48-L86) could be get rid of and replaced with a single command `pytest`. I assume that it is taken from `lightkde`, where I was playing with GitHub action and wanted to have a coverage badge without using any external service. If you do not mind using an external service then just get the coverage from one of those.

TTitscher:
Another hint, but no solution, because of the exit code 2...
> Exit code 2 signifies invalid usage of some shell built-in command. Examples of built-in commands include alias, echo, and printf.

TTitscher:
Can you temporarily remove the coverage stuff and just run pytest? Maybe one test fails and the redirection to stderr hides that... Hmm, but that would not trigger exit code 2, I guess :confused:

aklawonn:
When I run pytest locally, all tests pass. It seems like something went wrong after I added type hints in forward_model.py. Maybe I remove these and try again. And yes Arpad, I took the code from lightkde ;) It worked well so far.

aklawonn:
That's weird. I don't know why type hints would break it...

TTitscher:
Well, the type hints could be a separate PR and be ignored for this one. Strange nonetheless.

rozsasarpi:
Don't forget to squash the commits when merging. :)

aklawonn:
> Don't forget to squash the commits when merging. :)

Will do ;) This is such a weird error though. I don't know, what exactly triggers it.

probeye/definition/forward_model.py
```diff
@@ -70,21 +70,21 @@ def __init__(
...
 
     @property
-    def input_channel_names(self):
+    def input_channel_names(self) -> list[str]:
         """Provides input_channel_names attribute."""
         return self.input_sensor_names + [*self.prms_def.values()]
 
     @property
-    def output_sensor_names(self):
+    def output_sensor_names(self) -> list[str]:
```
rozsasarpi, line 83: should not this be `List[str]`? (`from typing import List`)

aklawonn: I will try this out...

rozsasarpi: I guess that `mypy` would have caught this issue before even reaching the testing phase. I think the following would prevent these issues happing in the future: `test` should be conditional on `lint_and_type_check` passing and the latter should have `mypy` checks.

aklawonn:
So, mypy-checks (static type checks) are now included, and the tests are not started unless the pre-commit checks (black, mypy) haven't encountered any issues. Still, this is ongoing work since the rest of the source code has to be adjusted to the new code style/format. What I'm doing here is mostly
1. to resize the docstrings to the new width of 88 characters
2. add the type hints and remove type-info in the docstrings

I will try to make some progress here every day, so that it doesn't take too long to merge this. So far, I have re-formatted `forward_model.py` and `sensor.py` from the `definition` sub-module.

rozsasarpi:
Thanks for the work so far, and for the summary of the status and your plan. Would you welcome/appreciate commits from others along your plans or you think it is easier to have the changes implemented by person?

aklawonn:
No, please feel free to participate and commit to this draft! I would definitely appreciate it.

PR #54, opened 2021-11-24
export to rdf

aklawonn:
In this branch, probeye's export-to-rdf feature should be developed. This feature should allow to export a well-defined inference problem to an rdf-file containing data triples relating to an ontology for parameter estimation problems (see probeye/parameter_estimation_ontology.owl), that describe the problem unambiguously.


aklawonn:
Most of the rdf export work is done at this point. This refers to the export of the problem definition. The inference data is not yet part of the export-framework. However, the export methods (and frankly also some other methods) are is not well tested yet. So I'm gonna work on the testing framework now.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # probeye changelog
 
+## 2.2.0 (2022-Mar-XX)
+### Changed
+- Before the evaluation of a likelihood function, the parameter domains are checked.
+### Added
+- Added an 'interface' submodule which contains a method 'export_to_rdf' that allows to print a defined problem to an rdf-file.
```
joergfunger, line 7: print -> export

docs/examples/plot_correlation_1D.py
```diff
@@ -150,13 +150,15 @@ def response(self, inp: dict) -> dict:
 problem.add_parameter(
     "std_noise",
     "likelihood",
+    domain="(0, +oo)",
```
joergfunger, line 153: looks like w weird composition of +oo. Even though I understand this, wouldn't it be better to pass a (lambda) function that evaluates to true (in the domain) or false outside. Thus, it would be much more flexible than pure intervals.

probeye/_setup_cfg.py
```diff
@@ -52,23 +55,38 @@ def version_constraint_free_dependencies(options_field: str) -> None:
...
+    config.read(setup_cfg)
     packages = config["options"][options_field]
 
     packages_without_version = version_constraint_free_packages(packages)
     setup_cfg_packages_without_version = "\n" + "\n".join(packages_without_version)
 
     config["options"][options_field] = setup_cfg_packages_without_version
 
-    with open("setup.cfg", "w") as configfile:
+    # just for testing during CI
```
joergfunger, line 74: what is that cfg file about? Not exactly sure that the test option does.

probeye/definition/inference_problem.py
```diff
@@ -2,6 +2,8 @@
 import copy
 from typing import Union, List, Optional, Callable, Tuple
 import copy as cp
+import urllib
```
joergfunger, line 5: what do you need urllib for here in this file??

probeye/definition/parameter.py
```diff
@@ -558,11 +623,11 @@ def check_consistency(self):
                 f"int or None, but found {type(self._dim)}."
             )
 
-        if (self._dim is not None) and (self._dim < 1):
-            raise ValueError(
-                f"Found value < 1 for ParameterProperties._dim! This attribute must be "
-                f"an integer >= 1, but found a value of {self._dim}."
-            )
+        # if (self._dim is not None) and (self._dim < 1):
```
joergfunger, line 626: not sure that that does, but either remove it, or uncomment

probeye/inference/scipy/likelihood_models.py
```diff
@@ -2187,8 +2164,10 @@ def loglike(
             if std_meas <= 0:
                 return worst_value
         else:
-            # consistent with tripy interface
-            std_meas = None
+            # in case of zero-residuals, a value of std_meas = 0 leads to a covariance
```
joergfunger, line 2167: why is that the residuals and not the value of the function

joergfunger (changes requested):
I will separately look into the owl first, but there are general questions related to the implementation (that is why I did not review everything yet). In addition, the RDF export seems to be mixed up with other changes.

probeye/parameter_estimation_ontology.owl
```diff
@@ -0,0 +1,938 @@
...
+        <rdfs:subClassOf rdf:resource="http://www.parameter_estimation_ontology.org#variable"/>
+    </owl:Class>
+    
+
+
+    <!-- http://www.parameter_estimation_ontology.org#single_experiment_data_set -->
+
+    <owl:Class rdf:about="http://www.parameter_estimation_ontology.org#single_experiment_data_set">
+        <rdfs:subClassOf rdf:resource="http://purl.obolibrary.org/obo/IAO_0000100"/>
+        <obo:IAO_0000115 xml:lang="en">A data set that is an aggregate of data recordings all of which have in common that they have been measured in a single experiment.</obo:IAO_0000115>
```
joergfunger, line 827: from my point of view, it is really difficult to define objectively, what a single experiment is. If you perform a test series (an experiment) on 5 different materials and perform 3 different tests each and use 5 specimens extracted from one bulk material (of the 5 materials), what is the number of experiments?

probeye/parameter_estimation_ontology.owl
```diff
@@ -0,0 +1,938 @@
...
+        <obo:IAO_0000115 xml:lang="en">A correlated Gaussian likelihood model the correlation of which is depending on space and time.</obo:IAO_0000115>
+        <obo:IAO_0000117 xml:lang="en">PERSON: Alexander Klawonn</obo:IAO_0000117>
+        <rdfs:label xml:lang="en">time-space correlated Gaussian likelihood model</rdfs:label>
+    </owl:Class>
+    
+
+
+    <!-- http://www.parameter_estimation_ontology.org#triple_constant -->
+
+    <owl:Class rdf:about="http://www.parameter_estimation_ontology.org#triple_constant">
```
joergfunger, line 847: the name is not very intuitive, and I'm also not sure why this is needed?

probeye/parameter_estimation_ontology.owl
```diff
@@ -0,0 +1,938 @@
...
+        <obo:IAO_0000115 xml:lang="en">A probability density function that is based on a uniform distribution.</obo:IAO_0000115>
+        <obo:IAO_0000117 xml:lang="en">PERSON: Alexander Klawonn</obo:IAO_0000117>
+        <rdfs:label xml:lang="en">uniform probability density function</rdfs:label>
+    </owl:Class>
+    
+
+
+    <!-- http://www.parameter_estimation_ontology.org#uninformative_probability_density_function -->
+
+    <owl:Class rdf:about="http://www.parameter_estimation_ontology.org#uninformative_probability_density_function">
```
joergfunger, line 887: Does an uninformative pdf (as a term) really exist, or is it in the implementation rather a specific distribution with specials choices on the parameters to make it noninformative. (I guess the term is rather noninformative than uninformative)

joergfunger (changes requested):
The ontology is quite complex, and I just started to look into the class definitions. From my point of view, it might be possible to remove some of the classes reducing complexity and allowing others to better understand the concepts.

CHANGELOG.md
aklawonn: Done

probeye/definition/inference_problem.py
aklawonn: Old imports. Not needed anymore. Removed.

probeye/definition/parameter.py
aklawonn: Forgot to delete. It is deleted now.

docs/examples/plot_correlation_1D.py
aklawonn: True. But this is the definition part of the problem where computing methods should be avoided. Providing a specific function might also lead to problems between numpy.arrays and torch.tensors when used by the solvers. The advantage of intervals is their simplicity and therefore easier definition in the ontology. Most application cases should be covered by the interval-option.

probeye/_setup_cfg.py
aklawonn: This file contains package-details like dependencies, version etc. This test covers the routine that checks if probeye still works with the latest versions of its dependencies.

probeye/definition/parameter.py
```diff
@@ -497,11 +520,52 @@ def __init__(self, prm_dict: dict):
         # check the given values
         self.check_consistency()
 
+    @staticmethod
```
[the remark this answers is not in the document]
aklawonn: See my comments above.

probeye/inference/scipy/likelihood_models.py
```diff
@@ -1384,21 +1363,19 @@ def loglike(
...
                 return worst_value
-            # this is the formula for the std. dev. of the sum of two normal dist.
-            std = np.sqrt(prms["std_model"] ** 2 + prms["std_measurement"] ** 2)
-        # the precision 'prec' is defined as the inverse of the variance, hence
-        # prec = 1 / sigma**2 where sigma denotes the standard deviation
-        prec = 1.0 / std ** 2.0
+            # this is the formula for the variance of the sum of two normal dist.
+            var = prms["std_model"] ** 2 + prms["std_measurement"] ** 2
+        else:
+            var = prms["std_model"] ** 2
```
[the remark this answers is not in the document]
aklawonn: No, the additive case. See [here](https://en.wikipedia.org/wiki/Sum_of_normally_distributed_random_variables).

probeye/parameter_estimation_ontology.owl
aklawonn: It's an own thing. The term "uninformative" is used for example [here](https://en.wikipedia.org/wiki/Prior_probability).

joergfunger: The term is used in the literature, but at the end it is always a specific distribution behind it. IMO when creating the distribution, there could be a class that creates a non-informative distribution (e.g. as an argument), the return value should however be the precise distribution that has been choosen. (Otherwise the description is not reproducible)

docs/examples/plot_correlation_1D.py
joergfunger: There are some applications, where the constraints actually include more than one parameter (e.g. they should be inside circle, on one side of a hyperplane, ..). That is tricky to model and requires additional effort. If it is just a function of the parameter dict (that you yould also provide/document (e.g. https://stackoverflow.com/questions/40828921/parsing-a-string-input-into-a-lambda-function-python) as a string, it is much less code, much easier to extend by someone and thus easier to maintain. As for the pyro, we we should have a short discussion.

probeye/inference/scipy/likelihood_models.py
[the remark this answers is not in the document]
joergfunger: so what is that std['model'] not sure that I understand that. I thought in the current implementation, the model uncertainty is zero. What exactly is that test for a the model std to be positive this should never be negative (I mean by definition)

docs/examples/plot_correlation_1D.py
aklawonn: I think this can be treated as an extension that is mostly independent from this PR. I opened a new issue #73 for that feature.

probeye/inference/scipy/likelihood_models.py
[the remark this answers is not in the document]
aklawonn: The model error is meant here as the error between the model response and the corresponding experimental data. A more elaborate name is the "model prediction error". The checks have been defined before parameter domains have been introduced, because it was possible that the sampler were proposing negative values for std-parameter, which would then lead to an error.

probeye/parameter_estimation_ontology.owl
aklawonn: So the uninformative prior is only relevant for maximum likelihood fits. An uninformative prior is equivalent with no prior defined. When performing a maximum likelihood fit, the user does not have to specify a prior. Internally, this is handled as an uninformative prior.

probeye/inference/scipy/likelihood_models.py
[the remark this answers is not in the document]
joergfunger: what are std_model and std_measurement?

probeye/parameter_estimation_ontology.owl
joergfunger: I would not fully agree. If you perform maximum likelihood, the prior is not used at all, i.e. it should not matter what distribution you are using. If you perform a full Bayesian approach, you have to specify a prior with a distribution function. There are different options discussed in literature, but they all have an underlying (flat) distribution that in particular depends on the scaling of the variable (so what could be considered a non-informative prior of a parameter x, it might not be a non-informative prior for a variable scaled by some factor e.g. y=1e6-6 x  (e.g. when changing from Pa to MPa).

PR #58, opened 2021-11-30
Dynesty solver

TTitscher:
Implement `dynesty` solver with "dynamic" and "static" nested sampling options. Remarks:
* ~~It would be nice to actually test the "dynamic" mode [here](https://github.com/BAMresearch/probeye/blob/dynesty_solver/tests/unit_tests/inference/dynesty_/test_solver.py#L54), but that takes minutes to complete, which I find too long for the test. @rozsasarpi are there dynesty options to speed that up?~~
* Nested samplers do not work with the logprior, but the prior quantile function. This is not always part of the `scipy` distributions, e.g. for the MVN. I specifically implemented the quantile function for the case of a uncorrelated MVN within the solver. Is that OK for now?
* The solver outputs `arviz.InferenceData` with only the posterior samples, which seems to be sufficient for our visualization. Is more information needed/preferred?

### DIFF
--- a/probeye/inference/dynesty_/solver.py
+++ b/probeye/inference/dynesty_/solver.py
@@ -201,6 +201,8 @@ class DynestySolver(ScipySolver):
         logger.info(f"Total run-time: {runtime_str}.")
         logger.info("")
 
+        logger.info("Resample weighted samples to equal samples for post")
+        logger.info("processing. Access the original dynesty results via  .raw_results")
         weights = np.exp(sampler.results.logwt - sampler.results.logz[-1])
         samples = dynesty.utils.resample_equal(sampler.results.samples, weights)
 
--- a/probeye/inference/dynesty_/solver.py
+++ b/probeye/inference/dynesty_/solver.py
@@ -74,8 +74,9 @@ class DynestySolver(ScipySolver):
                 i, j = np.nonzero(scale)
                 assert np.all(i == j)
 
-                for i in range(len(loc)):
-                    ppf.append(norm.ppf(q=x[i], loc=loc[i], scale=scale[i, i]))
+                mvn_ppf = norm.ppf(q=x, loc=loc, scale=np.diagonal(scale))
+                ppf += list(mvn_ppf)
+
         return ppf
 
     def get_summary(self, posterior_samples: np.ndarray) -> dict:
--- a/probeye/inference/dynesty_/solver.py
+++ b/probeye/inference/dynesty_/solver.py
@@ -1,0 +1,211 @@
+# standard library imports
+from typing import TYPE_CHECKING
+import time
+import random
+import contextlib
+
+# third party imports
+import numpy as np
+import emcee
+import arviz as az
+from loguru import logger
+from tabulate import tabulate
+
+# local imports
+from probeye.subroutines import pretty_time_delta
+from probeye.subroutines import check_for_uninformative_priors
+from probeye.inference.scipy_.solver import ScipySolver
+from probeye.subroutines import stream_to_logger, print_dict_in_rows
+
+# imports only needed for type hints
+if TYPE_CHECKING:  # pragma: no cover
+    from probeye.definition.inference_problem import InferenceProblem
+
+
+class DynestySolver(ScipySolver):
+    """
+    Provides emcee-sampler which is a pure-Python implementation of Goodman & Weare’s
+    Affine Invariant Markov chain Monte Carlo (MCMC) Ensemble sampler. For more
+    information, check out https://emcee.readthedocs.io/en/stable/.
+    """
+
+    def __init__(
+        self, problem: "InferenceProblem", seed: int = 1, show_progress: bool = True
+    ):
+        """See docstring of ScipySolver for information on the arguments."""
+        logger.debug("Initializing EmceeSolver")
+        # check that the problem does not contain a uninformative prior
+        check_for_uninformative_priors(problem)
+        # initialize the scipy-based solver (ScipySolver)
+        super().__init__(problem, seed=seed, show_progress=show_progress)
+
+    def emcee_summary(self, posterior_samples: np.ndarray) -> dict:
+        """
+        Computes and prints a summary of the posterior samples containing mean, median,
+        standard deviation, 5th percentile and 95th percentile. Note, that this method
+        was based on code from the taralli package: https://gitlab.com/tno-bim/taralli.
+
+        Parameters
+        ----------
+        posterior_samples
+            The generated samples in an array with as many columns as there are latent
+            parameters, and n rows, where n = n_chains * n_steps.
+
+        Returns
+        -------
+            Keys are the different statistics 'mean', 'median', 'sd' (standard
+            deviation), 'q05' and 'q95' (0.05- and 0.95-quantile). The values are
+            dictionaries with the parameter names as keys and the respective statistics
+            as values.
+        """
+
+        # used for the names in the first column
+        var_names = self.problem.get_theta_names(tex=False, components=True)
+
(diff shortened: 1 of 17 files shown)


joergfunger (changes requested):
I'm somehow confused with the title and the section dynesty suggesting a nested sampling procedure whereas the implementation uses emcee as a MCMC sampler?

tests/unit_tests/inference/dynesty_/test_solver.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+
+        # run the emcee solver with deactivated output
+        logging.root.disabled = True
+        dynesty_solver = DynestySolver(problem, show_progress=False, seed=6174)
+        _ = dynesty_solver.run_mcmc(n_walkers=20, n_steps=200, vectorize=False)
+        # summary = run_emcee_postprocessing(problem, emcee_sampler,
+        #                                    show_progress=True)
+        # sample_means = summary['mean']
+        # for mean, mean_true\
```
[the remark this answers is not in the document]
TTitscher: >This pull request is still a work in progress
Draft pull requests cannot be merged.

I opened that pull request only to let people know, I am working on it, in case someone else implements a dynesty solver. So nothing is finished -- not even started yet. :)

aklawonn:
Hi Thomas, thanks a lot for setting up the dynesty-solver so quickly! It looks pretty good to me. The only things I just added were some `run_dynesty` arguments in the integration tests which were missing, and I did the version increase from 1.0.14 to 1.0.15

aklawonn (approved):

probeye/inference/dynesty_/solver.py
```diff
@@ -0,0 +1,218 @@
...
+            )
+        end = time.time()
+        runtime_str = pretty_time_delta(end - start)
+        logger.info(f"Total run-time: {runtime_str}.")
+        logger.info("")
+        logger.info("Summary of sampling results")
+
+        self.raw_results = sampler.results
+
+        weights = np.exp(sampler.results.logwt - sampler.results.logz[-1])
```
joergfunger, line 206: So is that an additional plot routine extraction routine that resamples the output into the standard unweighted space? That also means that the results plotted are not fully the results obtained in the sampler, right?

tests/unit_tests/inference/dynesty_/test_solver.py
```diff
@@ -0,0 +1,62 @@
...
+        y_true = true["a"] * x_test + true["b"]
+        y_test = np.random.normal(loc=y_true, scale=true["sigma"])
+        problem.add_experiment(
+            f"Tests", fwd_model_name="LinRe", sensor_values={"x": x_test, "y": y_test}
+        )
+
+        # run the dynesty solver with deactivated output
+        logging.root.disabled = True
+        dynesty_solver = DynestySolver(problem, show_progress=False, seed=6174)
+        dynesty_solver.run_dynesty("static", nlive=50)
```
joergfunger, line 54: Shouldn't we use the dynamic version as "standard". At least from the literature is seems to yield more accurate results than the static version.

tests/unit_tests/inference/dynesty_/test_solver.py
```diff
@@ -0,0 +1,62 @@
...
+
+        sample_means = dynesty_solver.summary["mean"]
+        for parameter, true_value in true.items():
+            self.assertAlmostEqual(sample_means[parameter], true_value, delta=0.01)
+
```
joergfunger, line 59: The main feature of the nested sampling approach is to compute the evidence, could we use that for the analytical example where the evidence is known (in VB I guess) to have that as a test case here as well?

joergfunger (changes requested):
Great work, just a couple of potential improvements or questions from my side.

tests/unit_tests/inference/dynesty_/test_solver.py
TTitscher: The default method in the `DynestySolver.run` is indeed "dynamic". In this test, however, it runs for several minutes vs only seconds in "static" mode. Maybe I find a way to tweak the arguments of the "dynamic" mode to speed it up.

tests/unit_tests/inference/dynesty_/test_solver.py
TTitscher: IMO we should check, if dynesty is implemented correctly (another story), but only if our wrapper uses dynesty correctly. For this, I'd argue that it is enough to (roughly) check the means.

probeye/inference/dynesty_/solver.py
TTitscher: The probeye post-processing (currently??) relies on arviz.InferenceData which does not support weights. So I see no other way rather ran this `resample_equal` method. The full (weighted) results are still accessible in the `raw_results` member for users that do not want to use the probeye post-processing.

probeye/inference/dynesty_/solver.py
```diff
@@ -0,0 +1,218 @@
...
+        ----------
+        theta
+            A numeric vector for which the ppf should be evaluated.
+            Which parameters these numbers refer to can be checked by calling self.
+            theta_explanation() once the problem is set up.
+
+        Returns
+        -------
+        ppf
+            The vector of ppfs for each prior distribution at theta.
```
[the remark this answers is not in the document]
TTitscher: I fixed the naming. Its the quantiles, so the output of the ppf.

probeye/inference/dynesty_/solver.py
```diff
@@ -0,0 +1,218 @@
...
+                # exceptions, if not.
+                loc = prms[f"loc_{prior.ref_prm}"]
+                scale = prms[f"scale_{prior.ref_prm}"]
+                x = prms[prior.ref_prm]
+
+                # Assume the MVN to be uncorrelated, only diagonal nonzero
+                i, j = np.nonzero(scale)
+                assert np.all(i == j)
+
+                mvn_ppf = norm.ppf(q=x, loc=loc, scale=np.diagonal(scale))
```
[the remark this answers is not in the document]
TTitscher: scipy.stats.multivariate_normal has no `ppf`, which makes sense as a (e.g.) 2D MVN would have a circle/ellipse as a quantile. 

But I checked the [dynesty documentation](https://dynesty.readthedocs.io/en/latest/quickstart.html#example-non-uniform-priors) and they provide another solution for the prior transform... I'll look into that. Thanks!

probeye/inference/dynesty_/solver.py
joergfunger: OK, maybe just add some link where this resampling is discussed so that people interested in this know where to start, but otherwise I totally agree.

tests/unit_tests/inference/dynesty_/test_solver.py
joergfunger: But how do we extract the model evidence (in case that is of interest to someone)?

TTitscher: Ok, we now log the evidence [here](https://github.com/BAMresearch/probeye/pull/58/commits/669a15ff33099219d1550b269c1c1d72b054bd48#diff-07bccec89d5b89bfd486d23f61f3808db0a0a0d89909f5a0ebcef27113337a39R210)

joergfunger: Maybe that should not be part of the discussion in this pull request, but this feature (computing the model evidence based on the free energy in VB) would be computed also in other inference engines. Thus, it might be an option to include this (e.g. by extending the arviz class) to store this result (in case it is computed). But as mentioned, maybe this is a different issue.

joergfunger (approved):

Issue #69, opened 2022-02-25
Restriction of the parameter range


SMattfeldt:
The issue should address the question of how to restrict the range of parameters that can be used by the various solvers.
In many cases, such as the diffusion problem, realistic parameter values are limited to a certain range of values. If the value range is not narrowed down, unrealistic results can occur.

A larger problem occurs when numerical iteration methods are needed to solve the forward model, such as the Newton method. With poor starting values, the problem occurred that the Newton method could not converge.

A solution idea would be to generate an “if-statement” in the “loglike” function that can be defined by the user. If the current parameter value is outside of the specified parameter range, "-inf" is returned. Otherwise, the same calculation of the “loglike” function is carried out as previously specified. An advantage of this integration would be that the execution of the forward model for the current parameter values would be omitted and thus runtime could be saved.

This could be implemented, for example, by defining boundaries for each parameter of the problem and accessing these boundaries in the “if-statement”.

joergfunger:
In a similar line, but in a slightly different approach, a forward model would implement a check_parameters function that is called in the loglike function before the forward model itself (or by the user before the evaluation of the forward model). It would be great if you could make a branch and propose a pull request for the change. Maybe @aklawonn could point to the locations in the code that have to be changed.

aklawonn:
Thanks for your feedback Sina, we all appreciate it! It's a good idea with the parameter ranges, and the good news is that it is already implemented to some extend in the export_to_rdf branch (see for example [here](https://github.com/BAMresearch/probeye/blob/export_to_rdf/probeye/definition/inference_problem.py), line 249). As you can see, each parameter is assigned a domain, which is equivalent to what you call range. However, no checks on this domain are implemented yet, so this is something you could do, if this is something you would like to do and if you have the time for it.

Here's what I would propose:

1. Create a branch from export_to_rdf
2. Implement a check_parameter_domains method of the class InferenceProblem in probeye/inference_problem.py, which for a given parameter vector theta checks if all given values are in the domains of each parameter. This method would return True or False depending on the check.
3. In probeye/inference/scipy/solver.py there is a method called loglike. This is the place where the parameters would be checked. The code line to add would look like `self.problem.check_parameter_domains(theta).` The check would happen before the for loop. If the check doesn't pass, the method would immediately return -np.inf.

SMattfeldt:
The pull request is open: #70

aklawonn:
The respective changes have been merged to the _export_to_rdf_ branch.

PR #70, opened 2022-03-02
check parameter domain

SMattfeldt:
In this pull request a method is implemented that checks the parameters domain.
The loglike-methods now use this method to proof whether the latent parameters are within the specified domain. If they are not in the domain the loglike-methodes return  "-inf". 
The corresponding issue: #69

### DIFF
--- a/probeye/definition/inference_problem.py
+++ b/probeye/definition/inference_problem.py
@@ -1303,3 +1303,24 @@ class InferenceProblem:
                     f"The globally defined experiment '{exp_name}' does not appear in "
                     f"any of the likelihood models!"
                 )
+    
+    def check_parameters_domain(self, theta):
+        """
+        Checks weather the given values of the latent parameters are within the
+        specified domain.
+
+        Parameters
+        ----------
+        theta
+            A numeric vector or tensor, which contains the current values of all latent parameters.
+
+        Returns
+        -------
+        True if all values are within the specified domain. Otherwise, False is returned.
+        """
+        for theta_idx, theta_name in enumerate(self.get_theta_names()):
+            if theta[theta_idx] < self._parameters[theta_name].domain[0][0]:
+                return False
+            elif theta[theta_idx] > self._parameters[theta_name].domain[0][1]:
+                return False
+        return True       
--- a/probeye/inference/scipy/solver.py
+++ b/probeye/inference/scipy/solver.py
@@ -195,6 +195,13 @@ class ScipySolver:
         ll
             The evaluated log-likelihood function for the given theta-vector.
         """
+        
+        # check whether the values of the latent parameters are within the specified domain.
+        # if they are outside of the domain, break calculating the log-likelihood function
+        # and return instead of it minus infinity.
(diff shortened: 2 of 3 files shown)


probeye/inference/scipy/solver.py
```diff
@@ -195,6 +195,13 @@ def loglike(self, theta: np.ndarray) -> float:
         ll
             The evaluated log-likelihood function for the given theta-vector.
         """
+        
+        # check whether the values of the latent parameters are within the specified domain.
+        # if they are outside of the domain, break calculating the log-likelihood function
+        # and return instead of it minus infinity.
+        if self.problem.check_parameters_domain(theta) == False:
+            return -np.inf
```
aklawonn, line 203: A more pythonic way to write this condition would be:
```suggestion
        if not self.problem.check_parameters_domain(theta):
            return -np.inf
```

probeye/definition/inference_problem.py
```diff
@@ -1303,3 +1303,24 @@ def check_problem_consistency(self):
...
+        Returns
+        -------
+        True if all values are within the specified domain. Otherwise, False is returned.
+        """
+        for theta_idx, theta_name in enumerate(self.get_theta_names()):
+            if theta[theta_idx] < self._parameters[theta_name].domain[0][0]:
+                return False
+            elif theta[theta_idx] > self._parameters[theta_name].domain[0][1]:
+                return False
+        return True
```
aklawonn, line 1326: Your code works perfectly for scalar parameters! You probably didn't know, but you can also define vector-parameters with 2 or more components. I'm suggesting the following modification to make it work with them too:
```suggestion
        for theta_name in self.get_theta_names():
            theta_dim = self._parameters[theta_name].dim
            theta_idx = self._parameters[theta_name].index
            for i in range(theta_dim):
                if theta[theta_idx + i] < self._parameters[theta_name].domain[i][0]:
                    return False
                elif theta[theta_idx + i] > self._parameters[theta_name].domain[i][1]:
                    return False
        return True
        
```

probeye/inference/torch/solver.py
```diff
@@ -395,6 +395,14 @@ def loglike(self, theta: th.Tensor):
             A vector of pyro.samples (i.e. tensors) for which the log-likelihood
             function should be evaluated.
         """
+        
+        # check whether the values of the latent parameters are within the specified domain.
+        # if they are outside of the domain, break calculating the log-likelihood function
+        # and return instead of it minus infinity.
+        
+        if self.problem.check_parameters_domain(theta) == False:
+            return -np.inf
```
aklawonn, line 404: Same thing as before.
```suggestion
        if not self.problem.check_parameters_domain(theta):
            return -np.inf
```

aklawonn (changes requested):
Thanks for your contribution, the code looks good! I just suggested small modifications. As soon as you incorporate those (or something equivalent), I will merge your pull request.

By the way, we use [black](https://pypi.org/project/black/) as our code style. That means for example max. line length of 88 characters.

aklawonn (approved):

PR #77, opened 2022-04-27
surrogate model

aklawonn:
Implements the code structure for using surrogate models (meta models, response surfaces) with probeye. This structure is up for discussion, so please feel free to comment and propose changes. In order to check the currently intended use of a surrogate model, check out [this](https://github.com/BAMresearch/probeye/blob/surrogate_model/tests/integration_tests/tests_without_correlation/test_surrogate_model.py) integration test.

### DIFF
--- a/probeye/definition/inverse_problem.py
+++ b/probeye/definition/inverse_problem.py
@@ -10,6 +10,7 @@ import numpy as np
 # local imports
 from probeye.definition.parameter import Parameters
 from probeye.definition.forward_model import ForwardModelBase
+from probeye.definition.surrogate_model import SurrogateModelBase
 from probeye.definition.likelihood_model import GaussianLikelihoodModel
 from probeye.subroutines import underlined_string, titled_table
 from probeye.subroutines import simplified_list_string
@@ -95,6 +96,13 @@ class InverseProblem:
         # the script forward_model.py); this dictionary is managed internally and should
         # not be modified directly
         self._forward_models = {}  # type: dict
+
+        # the following dict contains the problem's surrogate models (note that
+        # surrogate models are optional, not every problem has surrogate models);
+        # the keys in this dict are the surrogate model names, while the values are the
+        # surrogate model objects (check out the script surrogate_model.py); this
+        # dictionary is managed internally and should not be modified directly
+        self._surrogate_models = {}  # type: dict
 
         # this dictionary contains the problem's likelihood models; as the other private
         # attributes above, it is managed internally and should not be modified directly
@@ -210,6 +218,11 @@ class InverseProblem:
     def forward_models(self) -> dict:
         """Access self._forward_models from outside via self.forward_models."""
         return self._forward_models
+
+    @property
+    def surrogate_models(self) -> dict:
+        """Access self._surrogate_models from outside via self.forward_models."""
+        return self._surrogate_models
 
     @property
     def experiments(self) -> dict:
@@ -639,6 +652,60 @@ class InverseProblem:
         # add the given forward model to the internal forward model dictionary under
         # the given forward model name
         self._forward_models[forward_model.name] = forward_model
+
+    def add_surrogate_model(self, surrogate_model: SurrogateModelBase):
+        """
+        Adds a surrogate model to the inverse problem. Note that multiple forward models
+        can be added to one problem.
+
+        Parameters
+        ----------
+        surrogate_model
+            Defines the surrogate model. Check out surrogate_model.py to see a template
+            for the surrogate model definition. The user will then have to derive her
+            own surrogate model from that base class. Examples can be found in the
+            package directory tests/integration_tests.
+        """
+
+        # log at beginning so that errors can be quickly associated
+        logger.debug(f"Adding surrogate model '{surrogate_model.name}'")
+
+        # check if all given model parameters have already been added to the inverse
+        # problem; note that the surrogate model can only be added to the problem after
+        # the corresponding parameters and the forward model have been defined
+        for prm_name in surrogate_model.prms_def:
+            self._parameters.confirm_that_parameter_exists(prm_name)
+            # check, if the type was set correctly; if it was not set yet, it will be
+            # set automatically here
+            if self._parameters[prm_name].type == "not defined":
+                self.change_parameter_type(prm_name, "model")
+            elif self._parameters[prm_name].type in ["prior", "likelihood"]:
+                raise ValueError(
+                    f"The parameter '{prm_name}' defined in forward model "
+                    f"'{surrogate_model.name}' was assigned the type "
+                    f"'{self._parameters[prm_name].type}' (it should be 'model')."
+                )
+
+        # check if the given name for the forward model has already been used
+        if surrogate_model.name in self._surrogate_models:
+            raise RuntimeError(
+                f"The name '{surrogate_model.name}' of the surrogate model you are "
+                f"trying to add to the problem has already been used for another"
+                f"surrogate model within the problem scope. Please choose another name."
+            )
+
+        # check if the surrogate model's forward model has already been added to the
+        # inverse problem
+        if surrogate_model.forward_model.name not in self._forward_models:
+            raise RuntimeError(
+                f"The surrogate model's forward model "
+                f"'{surrogate_model.forward_model.name}' has not been added to the "
+                f"problem yet!"
+            )
+
+        # add the given surrogate model to the internal surrogate model dictionary under
+        # the given surrogate model name
+        self._surrogate_models[surrogate_model.name] = surrogate_model
 
     # =============================================================== #
     #                   Experiments related methods                   #
--- a/probeye/definition/surrogate_model.py
+++ b/probeye/definition/surrogate_model.py
@@ -1,0 +1,32 @@
+# local imports
+from probeye.definition.forward_model import ForwardModelBase
+
+
+class SurrogateModelBase(ForwardModelBase):
+    """
+    Base class for a surrogate model, i.e., a forward model that approximates another
+    (typically computationally more expensive) forward model.
+
+    Parameters
+    ----------
+    name
+        The name of the surrogate model. Must be unique among all surrogate model's
+        names within a considered InverseProblem.
+    forward_model
+        The forward model object that the surrogate model approximates.
+    """
+
+    def __init__(self, name: str, forward_model: ForwardModelBase):
+
+        super().__init__(name)
+
+        # the surrogate model has access to the forward model it approximates; this
+        # forward model can be called (evaluating its response) from the surrogate model
+        # via self.forward_model.response(inp) where inp contains the input dictionary
+        self.forward_model = forward_model
+
+    def fit(self):
+        """
+        Prepares the surrogate model by approximating the forward model in some way.
+        """
+        pass
--- a/probeye/definition/__init__.py
+++ b/probeye/definition/__init__.py
@@ -2,6 +2,7 @@
 from probeye.definition import inverse_problem
 from probeye.definition import inference_problem
 from probeye.definition import forward_model
+from probeye.definition import surrogate_model
 from probeye.definition import likelihood_model
 from probeye.definition import parameter
 from probeye.definition import prior
--- a/tests/integration_tests/tests_without_correlation/test_surrogate_model.py
+++ b/tests/integration_tests/tests_without_correlation/test_surrogate_model.py
@@ -1,0 +1,176 @@
+# standard library imports
+import unittest
+import time
+
+# third party imports
+import numpy as np
+import matplotlib.pyplot as plt
+
+# local imports (problem definition)
+from probeye.definition.inverse_problem import InverseProblem
+from probeye.definition.forward_model import ForwardModelBase
+from probeye.definition.surrogate_model import SurrogateModelBase
+from probeye.definition.sensor import Sensor
+from probeye.definition.likelihood_model import GaussianLikelihoodModel
+
+
+class TestProblem(unittest.TestCase):
+    def test_surrogate_model(self, plot: bool = False):
+        """
+        Demonstrates how to use a surrogate model. Still under discussion.
+
+        Parameters
+        ----------
+        plot
+            Triggers a plot of the data used for calibration.
+        """
+
+        # ============================================================================ #
+        #                              Set numeric values                              #
+        # ============================================================================ #
+
+        # 'true' value of a, and its normal prior parameters
+        m_true = 2.5
+        mean_m = 2.0
+        std_m = 1.0
+
+        # 'true' value of b, and its normal prior parameters
+        b_true = 1.7
+        mean_b = 1.0
+        std_b = 1.0
+
+        # 'true' value of additive error sd, and its uniform prior parameters
+        sigma = 0.5
+        low_sigma = 0.0
+        high_sigma = 0.8
+
+        # the number of generated experiment_names and seed for random numbers
+        n_tests = 50
+        seed = 1
+
+        # ============================================================================ #
+        #                           Define the Forward Model                           #
+        # ============================================================================ #
+
+        class ExpensiveModel(ForwardModelBase):
+            def interface(self):
+                self.parameters = ["m", "b"]
+                self.input_sensors = Sensor("x")
+                self.output_sensors = Sensor("y", std_model="sigma")
+
+            def response(self, inp: dict) -> dict:
+                x = inp["x"]
+                m = inp["m"]
+                b = inp["b"]
+                time.sleep(1.0)
+                return {"y": m * x + b}
+
+        # ============================================================================ #
+        #                          Define the Surrogate Model                          #
+        # ============================================================================ #
+
+        class SurrogateModel(ExpensiveModel, SurrogateModelBase):
+            """
+            The inheritance from ExpensiveModel 'copies' the interface-method from
+            ExpensiveModel (the surrogate model should have the same interface as the
+            forward model). The inheritance from SurrogateModelBase is required to
+            assign a forward model to the surrogate model, see surrogate_model.py.
+            """
+
+            def response(self, inp: dict) -> dict:
+                x = inp["x"]
+                m = inp["m"]
+                b = inp["b"]
+                return {"y": m * x + b}
+
+        # ============================================================================ #
+        #                         Define the Inference Problem                         #
+        # ============================================================================ #
+
+        # initialize the inverse problem with a useful name
+        problem = InverseProblem("Using a surrogate model")
+
+        # add all parameters to the problem
+        problem.add_parameter(
+            "m",
+            "model",
+            tex="$m$",
+            info="Slope of the graph",
+            prior=("normal", {"mean": mean_m, "std": std_m}),
+        )
+        problem.add_parameter(
+            "b",
+            "model",
+            info="Intersection of graph with y-axis",
+            tex="$b$",
+            prior=("normal", {"mean": mean_b, "std": std_b}),
+        )
+        problem.add_parameter(
+            "sigma",
+            "likelihood",
+            domain="(0, +oo)",
+            tex=r"$\sigma$",
+            info="Standard deviation, of zero-mean additive model error",
+            prior=("uniform", {"low": low_sigma, "high": high_sigma}),
+        )
+
+        # add the forward model to the problem
+        forward_model = ExpensiveModel("ExpensiveModel")
+        problem.add_forward_model(forward_model)
+
+        # add the forward model to the problem
+        surrogate_model = SurrogateModel("FastModel", forward_model=forward_model)
+        problem.add_surrogate_model(surrogate_model)
+
+        # ============================================================================ #
+        #                    Add test data to the Inference Problem                    #
+        # ============================================================================ #
+
+        # data-generation; normal likelihood with constant variance around each point
+        np.random.seed(seed)
+        x_test = np.linspace(0.0, 1.0, n_tests)
+        y_true = surrogate_model.response(
+            {surrogate_model.input_sensor.name: x_test, "m": m_true, "b": b_true}
+        )[surrogate_model.output_sensor.name]
+        y_test = np.random.normal(loc=y_true, scale=sigma)
+
+        # add the experimental data
+        problem.add_experiment(
+            f"TestSeries_1",
+            fwd_model_name="ExpensiveModel",
+            sensor_values={
+                forward_model.input_sensor.name: x_test,
+                forward_model.output_sensor.name: y_test,
+            },
+        )
+
+        # plot the true and noisy data
+        if plot:
+            plt.scatter(x_test, y_test, label="measured data", s=10, c="red", zorder=10)
+            plt.plot(x_test, y_true, label="true", c="black")
+            plt.xlabel(forward_model.input_sensor.name)
+            plt.ylabel(forward_model.output_sensor.name)
+            plt.legend()
+            plt.tight_layout()
+            plt.draw()  # does not stop execution
+
+        # ============================================================================ #
+        #                           Add likelihood model(s)                            #
+        # ============================================================================ #
+
+        # add the likelihood model to the problem
+        problem.add_likelihood_model(
+            GaussianLikelihoodModel(
+                prms_def="sigma",
+                experiment_name="TestSeries_1",
+                model_error="additive",
+                name="SimpleLikelihoodModel",
+            )
+        )
+
+        # give problem overview
+        problem.info()
+
+
+if __name__ == "__main__":
+    unittest.main()


probeye/definition/inverse_problem.py
```diff
@@ -211,6 +219,11 @@ def forward_models(self) -> dict:
         """Access self._forward_models from outside via self.forward_models."""
         return self._forward_models
 
+    @property
+    def surrogate_models(self) -> dict:
```
joergfunger, line 223: why does the surrogate model has to be stored at the inference problem? An alternative would be
```
my_metamodel = gp_forwardmodel(orig_forward_model, additional_parameters_of_gp)
```

and then the user can decide on either using the metamodel
```
problem.add_forward_model(my_metamodel)
```
or the original forward model
```
problem.add_forward_model(orig_forward_model)
```

atulag0711:
Thanks Alex. The integration example looks nice and makes sense to me. However, fit method in probeye context would be difficult. I will complete my implementation, push here and then we can discuss.

mdbTNO:
Thanks for this. In my opinion the structure you propose makes a lot of sense, i.e. consider the surrogate model as a separate class and as model object. One remark is that we also have different ways to sample new datapoints to generate the surrogate. How would you connect the surrogate with the sampling methods?

joergfunger:
I would think the sampling could be done separately. So we would either have a sample_from_prior functionality, or any other sampler that takes as input a parameter estimation problem and returns the samples. Then these samples would be passed to the metamodel, which would then evalute the samples and build the metamodel. You could certainly perform the computation of the outputs (so the forward model output that is to be metamodeled) as a separate independent step, however doing it inside the metamodel allows for an adaptation (so if the metamodels requires new samples, they could just be added whenever the model is evaluated).

probeye/definition/surrogate_model.py
```diff
@@ -1,6 +1,15 @@
+import numpy as np
+from typing import Callable
+
 # local imports
+from probeye.definition.inverse_problem import InverseProblem
 from probeye.definition.forward_model import ForwardModelBase
 
+# Harlow imports
+from harlow.sampling import Sampler
```
joergfunger, line 9: Should that be in a very general definition of the interface (rather than in a specific implementation using harlow).

probeye/definition/surrogate_model.py
```diff
@@ -30,3 +39,293 @@ def fit(self):
...
+    of expensive forward models by bridging `probeye` with `harlow`. The
+    main objective is to reduce the manual effort required for surrogating
+    by eliminating the need to re-specify the problem when building the
+    surrogate.
+
+    """
+
+    def __init__(
+        self,
+        problem: InverseProblem,
```
joergfunger, line 66: We do we need that complex interface? In particular, why do we need the inverseProblem? What about the sampler, shouldn't the samples directly be passed?

probeye/definition/surrogate_model.py
```diff
@@ -30,3 +39,293 @@ def fit(self):
...
+        forward_model: ForwardModelBase,
+        sampler: Sampler,
+        surrogate_model: Surrogate,
```
joergfunger, line 69: This is also Harlow specific? How would we integrate different surrogate models (apart from Harlow)?

probeye/definition/surrogate_model.py
```diff
@@ -30,3 +39,293 @@ def fit(self):
...
+        self.evaluation_metric = evaluation_metric
+        self.verbose = verbose
+        self.run_name = run_name
+
+        # Find the experiment associated with the expensive forward model
+        # and get sensor information
+        # TODO: Implement check that the expensive forward model
+        #  has been added to the problem
+        self.input_sensor_data = dict()
+        for exp_name, exp in self.problem.experiments.items():
```
joergfunger, line 100: Why isn't all this information stored in the forward model?

joergfunger (changes requested):
Thanks for the integration. Can we somehow decouple the specific implementations from harlow from the general surrogate interface and put the specific implementations for harlow into a surrogate subdirectory? This way other metamodels could be integrated as well.

probeye/definition/inverse_problem.py
```diff
@@ -451,6 +482,35 @@ def get_parameters(self, theta: np.ndarray, prm_def: dict) -> dict:
...
+        prm_def
+            Defines which constants to extract. The keys of this dictionary are the
+            global parameter names, while the values are the local parameter names. In
+            most cases global and local names will be identical, but sometimes it is
+            convenient to define a local parameter name, e.g. in the forward model.
+
+        Returns
+        -------
+        prms
+            Contains <local parameter name> : <(global) parameter value> pairs. If a
```
joergfunger, line 503: not sure of I understand that, there is a local and global name, but the value is the same (so neither local or global)

probeye/metamodeling/initial_sampling.py
```diff
@@ -0,0 +1,175 @@
...
+        Returns
+        -------
+        sample_array
+            The sampled latent parameter vectors. Each row corresponds to a single
+            latent parameter vector.
+        """
+
+        # make sure that all parameters are one-dimensional; it is not straight forward
+        # how to do LHS for general multivariate parameters
+        for prm_name in self.problem.latent_prms:
```
joergfunger, line 63: Should we loop over the latent_prms of the inverse problem, or rather of the original forward model? In case of a single forward model, that is identical, but in case of multiple models, I don't think we should surrogate all models in parallel but rather split that into surrogates for each model.

probeye/metamodeling/initial_sampling.py
```diff
@@ -0,0 +1,175 @@
...
+    utilizing latin hypercube sampling. The generate samples are intended for training
+    of a surrogate model.
+
+    Parameters
+    ----------
+    problem
+        The considered inverse problem.
+    """
+
+    def __init__(self, problem: InverseProblem):
```
joergfunger, line 29: The design question if we surrogate the inverse problem, or each individual forward model. I would opt for the latter. The inverse problem will only provide in a somehow automatic way the bounds.

joergfunger (changes requested):
I did not review the complete code, but I realized one main design decision that I would like to dicuss. You are surrogating the inference problem (with all parameters), but shouldn't we surrogate each forward model and only use the inference problem to define bounds/distributions of the parameters?

JanKoune:
Hi, thank you for the feedback. It may be best to provide a brief description of the changes and the intended layout of the sampling and surrogating interface before addressing your individual comments:
1. Surrogating and sampling are now separated by defining the `SurrogateBase` and `SamplerBase` classes, from which the user is expected to derive their own implementation.
2. Both base classes should be as simple as possible for flexibility to allow for different approaches to surrogating and sampling. E.g. in a simple case we may want to draw LHS samples from the parameters and manually fit a surrogate model which we then use to define a forward model, or in a more complex case we may want to pass the surrogate model to the sampler to perform iterative fitting and sampling.
3. The main use of the `SurrogateBase` class is to copy the interface of a given forward model, and provide a template for surrogate models by implementing "empty" methods that are typically useful for surrogating (e.g. fit and predict). 
4. The `SamplerBase` is meant to read the inverse problem and forward model definitions and extract the information that is needed by the sampler (e.g. parameter priors and bounds, and input/output shapes). Indeed it would be best to do the surrogating for each forward model and not for the entire inverse problem. This could be enforced by making the forward model an input argument to `SamplerBase`, ensuring that a sampler is tied to a specific forward model. 
5. Most of the functionality for getting the problem and forward model information is not yet implemented in the base class. I will try to get it working asap.
6. The base classes are now decoupled from harlow.

examples/example_surrogate_hartman6D.py
```diff
@@ -0,0 +1,342 @@
...
+
+# =========================================================================
+# Create surrogate model
+# =========================================================================
+n_params = 4
+list_params = [[i for i in range(n_params)]] * len(sensor_names) * len(t_vec)
+
+# Kwargs to be passed to the surrogate model
+surrogate_kwargs = {
+    "training_max_iter": N_train_iter,
```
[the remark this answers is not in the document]
JanKoune: This is an argument expected by some GPyTorch based GP surrogates and defines the maximum number of training iterations, which usually contain the following steps ([from this example](https://docs.gpytorch.ai/en/stable/examples/01_Exact_GPs/Simple_GP_Regression.html)):

Zero all parameter gradients
Call the model and compute the loss
Call backward on the loss to fill in gradients
Take a step on the optimizer

Sklearn's GaussianProcessRegression has a similar argument to specify the number of restarts of the optimizer.

examples/example_surrogate_hartman6D.py
```diff
@@ -0,0 +1,342 @@
...
+
+# Define the surrogate model
+surrogate_model = VanillaGaussianProcess(**surrogate_kwargs)
+
+# Probeye's latin hypercube sampler
+lhs_sampler = LatinHypercubeSampler(problem)
+
+# An iterative sampler. Here we pass the surrogate ForwardModel directly to the sampler. However, it is
+# also possible to pass a surrogate model that will be included in a forward model after fitting.
+harlow_sampler = HarlowSampler(problem, forward_model, LatinHypercube, surrogate_model)
```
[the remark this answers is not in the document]
JanKoune: It is the [`harlow` implementation of LHS](https://github.com/TNO/harlow/blob/12-fix-bugs-in-imports/harlow/sampling/random_sampling.py). I will try to improve the documentation and fix the docstrings asap.

examples/example_surrogate_hartman6D.py
```diff
@@ -0,0 +1,342 @@
...
+harlow_sampler.sample(
+    n_initial_point=n_init,
+    n_new_point_per_iteration=n_point_per_iter,
+    n_iter=n_iter,
+    stopping_criterium=stopping_criterium,
+)
+harlow_sampler.fit()
+
+# Define the surrogate forward model
+forward_surrogate_model = HarlowSurrogate(
```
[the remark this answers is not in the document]
JanKoune: The `harlow_sampler.fit()` in line 224 is unecessary in this case, and simply refits the surrogate model using the final training samples and target points contained in `harlow_sampler` (drawn previously during the call to `harlow_sampler.sample(...)`. I will remove it in the next update.

probeye/metamodeling/initial_sampling.py
JanKoune: The newer version of `LatinHypercubeSampler` in `sampling.py` (which I simply copied from this implementation by Alex) is derived from `SamplerBase`. We could ensure that the sampling and surrogating is done for each individual forward model by specifying the forward model as an input argument in `SamplerBase`.

E.g. [this part](https://github.com/BAMresearch/probeye/blob/surrogate_model/probeye/metamodeling/sampling.py#L22):
```Python3
class SamplerBase:
    """
    Base class for probeye samplers
    """

    def __init__(self, problem: InverseProblem):
```

would become:

```Python3
class SamplerBase:
    """
    Base class for probeye samplers
    """

    def __init__(self, problem: InverseProblem, forward_model: ForwardModelBase):
```

probeye/definition/surrogate_model.py
JanKoune: Done, I have separated the general definition of the interface (`SamplerBase`) from the particular implementation for harlow (`HarlowSampler(SamplerBase)`).

probeye/metamodeling/sampling.py
```diff
@@ -1,47 +1,199 @@
...
+from probeye.inference.scipy.priors import translate_prior
+from probeye.subroutines import len_or_one
 
-# Harlow imports
-from harlow.sampling import Sampler
-from harlow.surrogating import Surrogate
+# external imports
+from harlow.sampling import Sampler as HarlowSamplerBase
+from harlow.surrogating import Surrogate as HarlowSurrogateBase
 from harlow.utils.helper_functions import latin_hypercube_sampling
```
danielandresarcones, line 20: I am not very fond of having external imports in the same script where the base class is defined. I would personally keep the base class separated from the derived ones, creating one new script for each group of samplers.

probeye/metamodeling/sampling.py
```diff
@@ -1,47 +1,199 @@
...
+        # the considered inverse problem
+        self.problem = problem
+
+        # the sampling happens before the problem is given to a solver; this means that
+        # the priors of the problem are merely descriptive and they have to be
+        # translated to have their respective computing routines
+        self.priors = copy.deepcopy(self.problem.priors)
+        for prior_template in self.problem.priors.values():
+            prm_name = prior_template.ref_prm
+            self.priors[prm_name] = translate_prior(prior_template)
```
danielandresarcones, line 37: I don't think the sampler should have to treat the priors itself, probably better in the interface.

probeye/metamodeling/sampling.py
```diff
@@ -65,8 +217,8 @@ def __init__(
         self,
         problem: InverseProblem,
         forward_model: ForwardModelBase,
-        sampler: Sampler,
-        surrogate_model: Surrogate,
+        sampler: HarlowSamplerBase,
+        surrogate_model: HarlowSurrogateBase,
```
danielandresarcones, line 221: Ideally a harlow sampler should be usable with a non-harlow surrogate, but it is not the case. Probably more a "harlow" thing than a "probeye" one.

probeye/metamodeling/surrogating.py
```diff
@@ -0,0 +1,430 @@
...
+        raise NotImplementedError
+
+    def response(self, inp: dict) -> dict:
+        """
+        Overrides the `response` method of the forward model.
+        """
+        raise NotImplementedError
+
+
+class HarlowSurrogate(SurrogateModelBase):
```
danielandresarcones, line 78: I would separate this from the base class.

probeye/metamodeling/surrogating.py
```diff
@@ -0,0 +1,430 @@
...
+        * Check out probeye's existing machinery for copying elements of the forward
+        model definition. Use that instead of `deepcopy`.
+
+    Parameters
+    ----------
+
+    """
+
+    def __init__(
+        self, name: str, surrogate: HarlowSurrogateBase, forward_model: ForwardModelBase
```
danielandresarcones, line 50: The base class should be general instead of forcing to input a harlow surrogate.
From my point of view, this base class should just coordinate the input and output of data to the surrogate model as well as its format.

danielandresarcones (changes requested):
I would suggest to develop the interface in such a way that separates completely the sampler and surrogate from the inverse and forward problem. The separated base classes work for that, but still require knowledge of the requirements of the `ForwardModel`. I woudl suggest something like [this](https://viewer.diagrams.net/?tags=%7B%7D&highlight=0000ff&edit=_blank&layers=1&nav=1&title=Interface.drawio#R7V1bc5s4FP41nsk%2BxIPA3B5tJ9ntTDubafb%2BkpFBttlicEFO7P76FVjiIoEhNpd4Q6fTmmMhxPnOTedI8kiZb%2FY%2FB3C7%2FuLbyB3Jkr0fKXcjWZYnhkn%2BiyiHI0XX5CNhFTj2kQRSwpPzA1GiRKk7x0ZhriH2fRc72zzR8j0PWThHg0Hgv%2BabLX03%2F9QtXCGB8GRBV6T%2B6dh4faQasp7Sf0HOas2eDDT6whvIGtM3CdfQ9l8zJOV%2BpMwD38fHT5v9HLkR8xhfjvc9lHybDCxAHq5zwx%2BPXyW4e1n%2BM5%2B6tms6j%2F%2BqX28ndHAv0N3RNx7Jmks6nG2jMeMDZYT2fRcNdLaBwcrxRspU2u6Ty1vsbwlpEpMw2uNb6DqrqJVFBocCQk17IJ9W9P%2F4OU6GADekn5krXn2KullCCyXk%2FC25HgkPHP4pi0CgMMLTjsjJCmI09ewn0p2LgkxPC%2F4%2BQtvytHUQMYsJL3tTUP7Sb2Gui5Y44W5Rl7OoMzpwZUou6Fv8SVRym76MyIKiV2l%2BZIy9dGzs8qzR9c9oaNvP28BfuGjzbEMMb37qlJsOUwTy3IifVP69RbjtGtgAhVvfC9lIUqDH2wDZjoVT1vQkepb%2FAgMHejG3unzw1vUxdrxV58Kx9APieBzfK3uynHum%2FIIC7BBvNz3a67vYkM%2Bo9b47PnTmk1ZLN%2FZdS4f4FGW29D1MfTWQ6fUD3Dhu5OV%2FQe4LinqN9BVv3KhR1IbcOvddP4ifrCBgq0gn9BAH%2FjeU%2BcbUdAVqyWizHo46vWjUaJ8hUY%2F3M%2FI3CAcH0oR%2Ba1L%2FdmBRhUTd82vqzAFrtM46coMFETSAWCV9pz6WfKBu9i0uFwgu98tBtIocUOGrs3Ghhxjz6TcSubbWjmt%2Fhgd%2FF409xND6xq5maz9wfpD2kKFAvg4YdIqUa%2FEU3Un7JLpN2jwy1gOO9AXucw0%2FwxCz0fiuC7ehs4jHBxIJnhGN8De0EScLNkTG0iqSBc0y0GLZjCxoMicMwJBEYZgUCQO7s3lhkAVh%2BITRhlCAIAJRYCXwyPOPMpFhKCVxSsxr%2Bsax7ajnWbiFFjFVn%2BNmd5OU8pUyYJI3AWtyI%2FIi8%2BUTZxQzRJ2Rv4RFc2msjlQy1jm5Buk1%2BRs1D%2FCc%2BAscQCfGEBGheUWR4MwCHxNXukhEWkT7tCpVywDFXKmJuNIa4EoZ4PIAeAuAa70DPhEA5yLiqXeoB33WHg7418Pf7B1%2FVcCfd%2FY309n8pw%2Fm8VVk2JMiCTfkhaI1FP0Zxtke32hNHjRBHqwAEWF43kTJsxtREAY3UKpTV%2BT3dQH2pYMHtFtBu3%2Bnbwho77Y2UfIB8FYA79%2FLmwLgmSTYgHjjiCdT8t4gB5JYO%2FkVr6PcjbTDjutgB4UD9m1gr%2FaNvSpCLyCNPHsaFSBTMDPJ0MDfeTayKafQ3sF%2Fkc%2B30liSJpTwd%2FTlGMT51ej6bp9pfXfIXDyiwCHvFVV%2FShjPkkn%2BLrBQDSyQnauKikgEyIXYeckXSYvYTG99jMQsE5jzeVlZ4qA6DpXelq1qcj3pWlVPZPqzQljoKcY9eaMLREFM51LOCwJBmPoZLpDLJlEZXcmIRoXKL%2Bisip9R3Wt30t19fbVjIkwL53Qso6RcXX%2BWdUs5cKFMJLVodou%2FXIaoHczErGs8pb4UsuBoeq4As2Yg4%2B1ri4iJadMvB6HMTVgKImhuiMG9hTbc4vjtPlpupbqa8rp2MHrawtgZvBL%2BNZNv4Ssspt57gUUV069DgaW%2BwVFqi8B7SbSoYr51KLC0CHjvuRZVTKjmFkMN5ZUW0e898aKKeVUxMJBuhmCg92BA73%2B1hSqmZYdgoL510GuLwLsJBsS07AXF94WhTtQCHVsaFrKsQQjeaYDABnBxgDBo%2F3XFBpqYmsuu9Z8mMUGy6v8jRQYnkk9v9vWKVsPXFy6zbc3Xs31GBVpf194nC5E%2FpsZr9cXghL8vQr09jRfThKm%2FH2BvCfYiD98t7GKSbzweD069BayLnHq3WIv5PQHok6XXqKbDPDnh6Npf%2BR5071MqX5z1yLD%2Fyl7QyqymM0Jamo2vDtmr2sXZY52yBlhtF2dlftWkxDvpusXZyaSqp5aLs5qYG7yK4iwT8h6Ks5cpp5iNu4rCanP8bqgWDjg9aa%2Byqok5sVP21Hbgxvfs39bRFkVi4eyHeEcgtXdsA788GZ1a7sIsarzeRePMKgDnmFW2hAaMsqtn1NGptTOlprhynQybXFTabDZrrrTZWX%2FK4qkLhUhV8tbX0M4045pZ0VGJFScyAw%2BZZjTeKR%2Bwzj1HVk%2BOi2%2BvmxNOQ44jaFZfxJRip%2FoSN83qi2L2ri%2FVaqD0pwZiOKOBMbN6b1UFZWKSwE9K%2Fihc13zY25BmiIEUZU7dyMs09PZVg6ljH6rRTHDerVZM6gb0nWiFfr5WaEAba4qZ%2FsnLnkm%2BNVKd4cbamorob1ORZOVouzoiRsj5avWwGbCjzYC169FmW6kMXQy9jynqYc9IjTmbfn3laF2MHYkC7Vz8DMPnYbNQa8D3XoI2xDzUqciocr8Ii3qSqXImUDkV89Dwhn2%2BKLxpfY8IZ69JBDGeyKVhxLkbRnTTHGciF5MLgVpOUBpiOHAVCUom0B9x94ghuu2rSHI2h9m17R4x5XdlfqM0DjByNlhRlJNWuGibX3Uhq9R2VyY4WaTStZE3jVaMvKmfY%2BTfOgU1AOex1NMzUL69aXQwATXFJRJX4XGYDn9Ej2OK6xuuwuM0h9nVeZy3rVIoS4WCBlKhVV6p1ElU5i9NraaTaCF%2FKez7lvgTm%2Br6Cmaly3tqeQ5gipPD%2Bz1x88SZj5KznAXpKdq1UaDiyQHc%2FJr1JdKswn0htm4upJOCccEBXFLRVFwqmIrzxrWxqbgpRu%2F%2FF25POEk2gdw3s5splp5lBjvIjbQZgTNFqbbCRn9WWOPLjLyHPTtgr1lIbcoGAzbT%2BJ9K6vlyyM4vrZRDwDZY9yGICrfWBShnr1nkT4kWempdFMVsRdZDla5nvw4fxY4YSqqubO1dVipAl04KAJHhD37wCgOb%2Fp4Nx%2BjyQ%2B2TnYDlewcb4OEtAHxYVXSsfRETFbU1Jop5hVMGtO6RWN2s50tEoO18lwDcuWvyhB82YL9I1JWVAmJO4irySImgXtvaagDEnMJVZIEa5HjHDBcn5p88MtgQPR5%2FAOgaPEPRhp2OPYNY4%2Fw9FMOXjHPwt8jLBdLxIYg2DNexfwC1our0MMVMaD1RqoLr%2FheXJ3JXvYBQqum1moy4k7PUJX3M31bXewE%2Be1fUWXMeDH0Lv8Mf3yzZ%2Fff7ww%2FpQZpqv9%2BKE75Iq4lxihi2gCGyXBiGUS9eiBEkZlbyl%2BSf6WwuyC5dpJMRyreu10mi8tIVO0XhfV70GzAgQOJySHrRWatykf1oIDwvxEkMzk%2BZjqq48ny11M%2FYyqEWMEptZsYLJL7aeO6MNw1CGeZ6azPeQoTFmcOgiaImgsJjjztVRTHkb0UV342GCT8Yxi8FPlvFSFDUrY6dtWu6LOKK6I8QRzmpmCJLSqJo7Fdy5avDm98SYZ47N0%2FlpKynlsEW5y1dBNyd1Z3ZqV69bKrkrLKhqucJCX88XlKGvlhGyGX6q9LH5ulvcyv3%2FwE%3D), where the `SurrogateAndSampler` interface controls the flow of information with the inverse problem and between sampler and surrogate, while acting as the `ForwardModel`. This would allow to add the necessary utilities in the interface instead of having to redefine them for every surrogate. More complex interactions such as adaptative samplers can be implemented as children. The data format between the interface and the individual wrappers should be decided  beforehand.

Additionally, the wrappers are to be implemented individually, allowing the user to adapt the inputs to their specific solver. If, for example, a surrogate needs information of a specific forward model, that can be requested in the wrapper and introduced in the initialization, encapsulating each of the implementations individually.

probeye/metamodeling/sampling.py
JanKoune: This is done for convenience initially, and it would be trivial to separate the specific implementations from the base class if necessary later.

probeye/metamodeling/sampling.py
JanKoune: Harlow offers an abstract surrogate class which can be used to define new surrogates that can be used with the harlow samplers. Since most samplers need access to the surrogate (for fitting and making predictions iteratively), I don't think its feasible to make them work with any surrogate.

probeye/metamodeling/surrogating.py
JanKoune: Indeed, thanks for catching this!

JanKoune:
Thank you for the suggestions @danielandresarcones. Some notes:

1. What is the benefit of creating surrogating and sampling wrapper base classes plus an interface class? Why not use the `__init__` method of `SamplerBase` to obtain all the information needed for surrogating and sampling from `InverseProblem` and `ForwardModel`?
2. Ideally, with the current structure of `SamplerBase` and `SurrogateBase` the user would not need any knowledge of the `InverseProblem` and `ForwardModel` internals since extracting the necessary information would be taken care of in the `__init__` of `SamplerBase`, although this is not yet implemented.
3. I like the idea of having the interface also act as the forward model. However, it seems that there is a lot of processing of the forward model that happens internally in probeye and could potentially cause problems if the forward model becomes too complex (which is why I opted to keep `SurrogateBase` as light as possible).

danielandresarcones:
1. My idea behind it is to separate functions and allow an easier integration with external generic samplers/surrogates, unless we decide to go always through harlow. Using the `__init__` from SamplerBase means that the interactions between surrogate and sampler are controled by the sampler, and that this one will have to store the information regarding datasets, formats and parameters, as well as implementing the extra utilities that we may need. If we want to use a surrogate with a previously generated dataset with no sampling at all, having to define a sampler seems counterintuitive. In the same way, derived samplers shouldn't be able to modify the utilities concerning the ForwardModel, such as comparing the output of the surrogate model or generating the covariance matrices.
2. The derived class would not have to modify it, but it will still be passed through the `__init__ `and accessible, as well as having to generate a copy of it. I consider encapsulating the sampler and surrogate from the inverse and forward problems the safest option.
3. That is a good point, although it is something to consider.

JanKoune:
I don't see anything in the current implementation tying us to Harlow or necessitating the use of a sampler to fit a surrogate model to an existing dataset. It seems to me that the two approaches are mostly functionally equivalent, with the main differences being:

- Having a separate interface class vs. having `SamplerBase` act as the interface
- Defining the forward model as part of the interface vs. defining it as part of the surrogate model

I do not have a strong preference for either approach.

examples/example_surrogate_hartman6D.py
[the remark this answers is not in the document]
joergfunger: 2 . Why does the sampler need the forward_model and the surrogate? I would imaging that this is aleady an adaptive procedure that creates samples, potentially evaluates the surrogate to identify the relevant points, samples new points and so forth. However, this would already build the surrogate, so I'm not sure what steps for is for (why not directly use the surrogate created in 1 as the forward model? I'm also not sure why we would need the problem? If this is for he initial sampling, I would rather create that initial set of points within probeye.

probeye/metamodeling/initial_sampling.py
joergfunger: I see the challenge, however it is also tricky to hide that. What do we do in case of a parametric prior (where e.g. the std. dev is again a parameter).

probeye/metamodeling/sampling.py
joergfunger: I think the challenge with the implementation is that we strictly separated between the problem definition and compute functions (such as evaluating the prior with e.g. a scipy pdf evaluation, or a pytorch pdf evaluation), i.e. the problem definition cannot perform a sampling from the prior a priori. Thus, we would somehow have to create a solver just for sampling the prior using e.g. LHS. Not sure what exactly would be the recommended option, I see two options
1. The sampler just get's the distributions and performs the (initial) sampling internally. This would mean to extract the distributions with all the relevant parameters from the inference problem.
2. We create a dummy sampler that is able to sample from the prior and extracts that samples as a first set of points to the surrogate. 
In both situations, the intervals for the variables have to be transferred to the surrogate sampling either by having a check_bounds in the inference problem, or by "copying" or extracting these information from inference problem.

probeye/metamodeling/sampling.py
joergfunger: Why exactly should that not work with other surrogates? But I somehow agree that the sampler and the surrogate are connected and maybe we should actually follow a suggestions from Daniel to put that into a same base class.

examples/example_surrogate_hartman6D.py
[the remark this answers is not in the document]
JanKoune: The `HarlowSampler` needs access to the "real" forward model to generate new pairs of sample points - observations, and the surrogate model to fit and evaluate the accuracy iteratively. The problem is passed so that we can obtain information on the parameters and bounds.

To avoid passing both a surrogate and a forward model to the sampler, we could restrict the sampler base class to only take a `SurrogateBase` derived object instead of any `surrogate_model`, since the `SurrogateBase` object already contains the real forward model that we want to surrogate. The previous code would look something like:

```Python3
surrogate_model = VanillaGaussianProcess(**surrogate_kwargs)
forward_surrogate_model = HarlowSurrogate("SurrogateModel", surrogate_model, forward_model)
harlow_sampler = HarlowSampler(problem, forward_surrogate_model, LatinHypercube)
harlow_sampler.sample(
    n_initial_point=n_init,
    n_new_point_per_iteration=n_point_per_iter,
    n_iter=n_iter,
    stopping_criterium=stopping_criterium,
)
```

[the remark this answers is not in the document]
joergfunger: As far as I see yesterdays discussion, we would have an (adaptive) sampler that gets a surrogate (to evaluate the surrogate for a decision on where the next point is) with the forward model to actually evaluate the model and a surrogate (that only has predict, update and fit). I guess this answers above discussion.

PR #88, opened 2022-07-31
solver reproducibility

aklawonn:
Sampling solver results are now reproducible. This is explicitly checked in corresponding tests.


joergfunger (commented):
Nice work, thanks, why did you delete the branch?

PR #104, opened 2023-07-31
103 output sensor order in forward model

danielandresarcones:
Added check for same sensor definition order in the scipy solvers beetween model and experiment response. If they are different, the experiment reponse dictionary is modified for the forward model, therefore it is only run during the first evaluation. Added test based on @ppp0l 's suggested MWE.


joergfunger (approved):
Looks good, thanks.

PR #114, opened 2024-08-26
Embedded bias initial implementation

### LINKED ISSUE #90: Avoid copy.deepcopy (restriction in FEniCS)
A lot of models we practically wish to calibrate with probeye are developed in FEniCS, as it is the main simulation tool in our group. Nevertheless, in FEniCS there is a restriction: we cannot _deepcopy_ a FEniCS function, which is very often used in a simulator. In probeye there are currently some places where the inference problem and its ingredients (mainly the forward model/simulator) are subjected to 'deepcopy'. This obviously raises failing errors.

I am not sure which reasons are considered for making deepcopy of the inference problem in probeye. Maybe it has been a safe way of passing the same problem to several solvers. If so, perhaps a better approach instead of making deepcopy would be to introduce methods like 'reset_problem', such that the problem and particularly the forward simulator behind it can be returned to their original default state. Maybe this method can be considered as a _must_ attribute of any forward simulator that a user is going to calibrate by probeye.

### LINKED ISSUE #107: [WIP] Inference with model bias
WIP of additional inference solver and likelihood models for bias identification.

danielandresarcones:
This is the initial implementation of embedded bias approaches to be integrated in probeye. It provides a working implementation for the computation of inverse problems with embedded bias.
- Currently the bias embedding must be implemented manually in the forward model. Ideally, this should not be the case and the bias should be specified in the parameters and variables description.
- The additions take care of the MCMC solver with stochastic forward model response and introduce four embedded likelihood formulations: ABC Moment matching, Global Moment Matching, Relative Global Moment Matching and Independent Normal. The details on the implementations will eventually be published.
- Kennedy and O'Hagan-type formulations are not implemented in this PR and have been eliminated from the branched code.
- Tests and one simple example for the documentation have been added. 
- `chaospy` has been added as a dependency. Probably should be an optional one, but kept as compulsory for now.
- `mypy` was updated to the latest version


danielandresarcones:
This update also implements the proposed solution of Issue #90 .

danielandresarcones:
Tests cover is not 100% on the new implementations as there probably will be changes in the definition of the bias in the near future. The lack of cover is precisely on that front, the checks that some parameter definitions are correctly imported to the models. Nevertheless, they can be tested if considered necessary.

docs/examples/bias.py
```diff
@@ -0,0 +1,264 @@
...
+# the model response. The model is defined as a class that inherits from the
+# `ForwardModelBase` class. The `interface` method is used to define the model's
+# parameters, input sensors, and output sensors. The `response` method is used to
+# calculate the model's response based on the input parameters. In this case, the model
+# is a simple linear regression model with slope `a` and bias scale `b`. The bias term
+# is modeled as a normal distribution with zero mean and standard deviation `b`. The
+# model response is calculated using PCEs with pseudo-spectral decomposition.
+
+
+class LinearModel(ForwardModelBase):
```
joergfunger, line 94: This is an elegant way of using the flexibility with the Forward model output to return directly a distribution rather then sensor output. But that means the ForwardModel is not compatible with other deterministic ForwardModels. Would it make sense to Create a wrapper for the StochasticForwardModels, that takes as input a deterministic forward model (as well as the definition of the embedding and the parameters of the PCE) and then wraps that pce expansion as a blackbox, anyone else could then easily add his/her model, and then you could also check, whether the Forward model send to the embedded solver is actually a stochastic forward model.

docs/examples/bias.py
```diff
@@ -0,0 +1,264 @@
...
+
+        m = np.repeat(m, len(x))
+        x = x.reshape(-1, 1)
+        m = m.reshape(-1, 1)
+
+        # define the distribution for the bias term
+        b_dist = chaospy.Normal(0.0, b)
+
+        # generate the polynomial chaos expansion
+        expansion = chaospy.generate_expansion(self.pce_order, bias_dist)
```
aradermacher: It should be b_dist instead of bias_dist

aradermacher (commented):
Thanks Daniel, 
I had problems running the example file, because of a typo (see comment)
Furthermore, during creating a new env I run into an error with python 3.13. Changing to 3.10 solved it.

docs/examples/bias.py
danielandresarcones: I missed that totally, thanks!

danielandresarcones:
Do you remember what whas the error with python 3.13? I think probeye had some problems with it at some point, not necessarily the embedded bias implementation only.

aradermacher:
> Do you remember what whas the error with python 3.13? I think probeye had some problems with it at some point, not necessarily the embedded bias implementation only.

Yes it was something with the required packages (but I didn't saved the message sorry). So I think it's a general problem of probeye and not with your implementation.

danielandresarcones:
This version is the first step towards #115. Main changes:

- Parallelization of walkers in emcee-based solvers.
- Simplified creation of new likelihoods (bias likelihoods were created in the new way). They can be defined without having to deal with the correlation features
- Fixed the bias implementations
- Solved copy/deepcopy issue from #90

docs/examples/bias.py
```diff
@@ -0,0 +1,261 @@
...
+        self.pce_order = 1
+
+    def interface(self):
+        self.parameters = ["a", "b"]
+        self.input_sensors = Sensor("x")
+        self.output_sensors = Sensor("y", std_model="sigma")
+
+    def response(self, inp: dict) -> dict:
+        x = inp["x"]
+        m = inp["a"]
```
joergfunger: the naming is not clear, (I guess m stands for mean, b for bias), but some short documentation would be nice

docs/examples/plot_correlation_1D.py
```diff
@@ -198,14 +198,16 @@ def response(self, inp: dict) -> dict:
 # the emcee solver, which is a MCMC-sampling solver. Let's begin with the scipy-solver:
 
 # this is for using the scipy-solver (maximum likelihood estimation)
-scipy_solver = MaxLikelihoodSolver(problem, show_progress=False)
-max_like_data = scipy_solver.run()
+# scipy_solver = MaxLikelihoodSolver(problem, show_progress=False)
```
joergfunger, line 201: why is that uncommented?

docs/examples/plot_correlation_1D.py
```diff
@@ -198,14 +198,16 @@ def response(self, inp: dict) -> dict:
...
+# max_like_data = scipy_solver.run()
 
 # %%
 # All solver have in common that they are first initialized, and then execute a
 # run-method, which returns its result data in the format of an arviz inference-data
 # object (except for the scipy-solver). Let's now take a look at the emcee-solver.
 
+problem.forward_models.pop("LinearModel")
```
joergfunger: Why is the model first removed and then added?

probeye/definition/inverse_problem.py
```diff
@@ -92,11 +92,14 @@ def __init__(
         # this dict is intended for storing the problem's experimentally measured data
         self.experiments = {}  # type: dict
 
-        # the following dict contains the problem's forward models; note that a single
+        # the following dicts contains the problem's forward models; note that a single
         # problem can have multiple forward models; the keys in this dict are the
         # forward model names, while the values are the forward model objects (check out
         # the script forward_model.py)
+        # the internal forward models list enables reusing the same problem in different
```
joergfunger, line 99: not sure I understand that, why do we need internal forward models? Why don't you define difference inference problems when using different solvers?

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+            return -np.inf
+
+        # Adjustable weights for the mean and std of the moment residuals
+        if not hasattr(self, "weight_mean"):
+            self.weight_mean = 1.0
+        if not hasattr(self, "weight_std"):
+            self.weight_std = 1.0
+
+        # Load the standard deviations and noise values
+        std_model, std_meas, stds_are_scalar = self.std_values(prms)
```
joergfunger, line 119: not sure I understand what that methods does? Could you add some docs?

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+            return -np.inf
+
+        # Load the standard deviations and noise values
+        std_model, std_meas, stds_are_scalar = self.std_values(prms)
+        variance = np.power(std_model, 2)
+        n = len(residual_vector)
+
+        # Calculate the intermediate statistics
+        mean_residual = np.mean(residual_vector)
+        variance_residual = np.var(residual_vector)
```
joergfunger: I think this is assuming the mean to be known, but since it is estimated, you have to divide by n-1 rather than n, use ddof=1 in this case

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+        mean_var = np.mean(np.square(response_vector[1]))
+        sample_variance = mean_var + variance_residual + variance
+        population_variance = (
```
joergfunger: I think the variable notation is not matching with the paper, it would be nice if that could be improved (e.g. population variance is defined differently in the paper in eq. 31)

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+
+        data = response_vector[0] + residual_vector
+
+        # Load the standard deviations and noise values
+        std_model, std_meas, stds_are_scalar = self.std_values(prms)
+        variance = np.power(std_model, 2)
+        n = len(residual_vector)
+
+        # Calculate the intermediate statistics
+        sigma_model_population = np.sqrt(
```
joergfunger: not sure I see this in eq. 42

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+            self.gamma**2
+            * np.square(residual_vector)
+            # + variance
+        )
+        sigma_model_sample = np.sqrt(np.square(response_vector[1]) + variance)
+        population_variance = np.var(np.divide(data, sigma_model_population)) + 1
+        sample_variance = np.var(np.divide(response_vector[0], sigma_model_sample)) + 1
```
joergfunger: not sure I would I see the similarity to eq. 47 in the paper

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+            ll += ((n - 1) / 2 - 1) * np.log(n * sample_variance / population_variance)
+        else:
+            raise NotImplementedError("Heteroscedastic noise not implemented")
+
+        return ll
+
+
+class RelativeScaledGlobalMomentMatchingModelError(EmbeddedLikelihoodBaseModel):
+    """
+    This class implements the relative global moment matching likelihood model.
```
joergfunger: not sure what the difference between the scaled global moment matching and the unscaled version is? Is there a reference in the paper for that likelihood?

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+        if np.isnan(response_vector).any():
+            return -np.inf
+
+        # Load the standard deviations and noise values
+        std_model, std_meas, stds_are_scalar = self.std_values(prms)
+        variance = np.power(std_model, 2)
+        n = len(residual_vector)
+
+        # Calculate the intermediate statistics
+        sigma_model_population = np.sqrt(
```
joergfunger: is that eq. 42?

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+            self.gamma**2
+            * np.square(residual_vector)
+            # + variance
+        )
+        sigma_model_sample = np.sqrt(np.square(response_vector[1]) + variance)
+        population_variance = (
+            np.var(np.divide(residual_vector, sigma_model_population)) + np.mean(1 - np.divide(variance, np.square(sigma_model_population)+variance))
+        )
```
joergfunger: not sure I can follow here with the equations from the paper

probeye/inference/bias/likelihood_models.py
```diff
@@ -0,0 +1,431 @@
...
+        variance = np.power(std_model, 2)
+        sigma_model_sample = np.sqrt(np.square(response_vector[1]) + variance)
+
+        ll = 0
+        if std_meas is not None:
+            variance += np.power(std_meas, 2)
+        if stds_are_scalar:
+            ll -= 0.5 * np.sum(
+                np.square(np.divide(residual_vector, sigma_model_sample))
+                + np.log(2 * np.pi * np.square(sigma_model_sample))
```
joergfunger: where is the -ny/2 from eq. 28? and where is the inverse in the last term?

probeye/inference/scipy/solver.py
```diff
@@ -105,7 +105,7 @@ def _translate_forward_models(self):
             forward_model.prepare_experimental_inputs_and_outputs()
 
             # finally, add the forward model to the problem
-            self.problem.forward_models[fwd_name] = forward_model
+            self.problem.internal_forward_models[fwd_name] = forward_model
```
joergfunger, line 108: see above not sure what these internal forward models are

probeye/inference/solver.py
```diff
@@ -31,7 +31,7 @@ def __init__(
     ):
 
         # attributes from arguments
-        self.problem = cp.deepcopy(problem)
+        self.problem = cp.copy(problem)
```
joergfunger, line 34: why is a copy here sufficient?

joergfunger (changes requested):
nice work, thanks. There are a couple of remarks or thinks that I have difficulties to relate to the current paper. It would be nice if the naming conventions in the paper and the source code are somehow overlapping, such that the paper serves as a scientific documentation to the source code.

probeye/inference/bias/likelihood_models.py
danielandresarcones: Updated

probeye/inference/bias/likelihood_models.py
danielandresarcones: See updated version

probeye/inference/bias/likelihood_models.py
danielandresarcones: See updated version

docs/examples/plot_correlation_1D.py
danielandresarcones: Forgotten from debugging, removed

docs/examples/plot_correlation_1D.py
danielandresarcones: Forgotten from debugging, removed

docs/examples/bias.py
danielandresarcones: Modified all the examples for consistent naming. Now "a" is the slope and "b" is the cross with the y-axis.

docs/examples/bias.py
danielandresarcones: Agree, it will be the case for the reimplementation of probeye. For now it remains as something the user must consider when building their forward models.

probeye/inference/scipy/solver.py
danielandresarcones: These are here to solve the copy/deepcopy problem. See #90 for full detail, but the problem is triggered when several solvers access the problem sequentially. FEniCS(X) models cannot be deepcopied, so we need to change them all to copies. Then, the forward model definition is overwritten in the solver initialization, which means that the next solvers will  see only the hull and throw an error. By assigning the translated forward model to the list "internal_forward_models" instead of overwriting "forward_models" and adapting the solvers accordingly, we still preserve the forward model definition after the first call, and is the "internal" one that is overwritten. The deepcopies are therefore not needed as the forward_model definition is still available after running the problem and a new "internal" forward model is created and evaluated from it.

probeye/inference/solver.py
danielandresarcones: See above.

probeye/inference/scipy/solver.py
joergfunger: I see, thanks. Just one add-on question. Is it required to store the these internal forward models on the level of the problem definition, or might it be sufficient to store them only locally in the solver, where these copies are actually required? In addition, the deep copy was done to prevent the copies to interact if they run in parallel, which as far as I understand here is not intended (the copy would result in identical objects e.g. for the function spaces that are shared between the different copied models). Wouldn't that mean we could just work on the original model, i.e. preventing parallel execution and in this case avoid any copy?

probeye/definition/inverse_problem.py
danielandresarcones: See below. It is so you can use a shallow copy to re-use the forward model and not needing to do the pop+readding trick or defining a totally new problem. Just for consecutive uses of the same forward model (relevant for reinitiating a run)

probeye/inference/bias/likelihood_models.py
danielandresarcones: Updated in the new version

probeye/inference/bias/likelihood_models.py
danielandresarcones: Updated in the new version

probeye/inference/bias/likelihood_models.py
danielandresarcones: Updated in the new version

probeye/inference/bias/likelihood_models.py
danielandresarcones: I removed all the likelihoods that are not being used in the paper, they were just variations that I tried at some point.

probeye/inference/bias/likelihood_models.py
danielandresarcones: Updated in the new version

probeye/inference/bias/likelihood_models.py
danielandresarcones: Do you mean extending the docstring of the method or writing a comment there in the application? It is a method from the base class, it is defined in ScipyLikelihoods

probeye/inference/scipy/solver.py
danielandresarcones: We could store them in the solver, but that means modifying the way the likelihood access to them as well (right now it reads from the problem). We could avoid the copy by creating new objects as long as we do not create a hull anymore, which means that definition and execution of the forward model would not be decoupled anymore. If that is still necessary is something we would have to discuss.

joergfunger: I guess this is something for the redesign of the interface, lt's tackle that in the subsequent redesign.

joergfunger (approved):
Thanks

